PR #1, opened 2015-03-06
Design updates

marcgurney:
Updated header styles, removed hero, updated type and image treatment for post lists.

screen shot 2015-03-06 at 10 31 51 am


elianekabkab:
yay! lgtm (g for great)

PR #4, opened 2017-07-17
js-feat-input-form

sampoonachot:
- Added info and darker buttons
- Added input states for success and error

### DIFF
--- a/client/components/ui-design-system/buttons/index.jsx
+++ b/client/components/ui-design-system/buttons/index.jsx
@@ -128,6 +128,35 @@ const UIButtons = () => (
       </div>
     </div>
 
+    <div className="row u-mb-2">
+      <div className="columns small-12">
+        <h2>Links</h2>
+      </div>
+    </div>
+    <div className="row u-mb-2">
+      <div className="columns small-12">
+        <h3>Base Styles</h3>
+          <a className="link">This is a normal styled link</a><br></br>
+          <a className="link link-is-underlined">This link is always underlined, even when not hovering</a><br></br>
+          <a className="link link-on-dark">This link is brighter, to be used against a blue background</a><br></br>
+          <a className="link link-is-disabled">This link is disabled, which prevents click and hover events like tooltips</a><br></br>
+          <a className="link link-is-not-allowed">This link is styled as disabled, but allows click and hover events like tooltips </a>
+      </div>
+    </div>
+    <div className="row u-mb-3">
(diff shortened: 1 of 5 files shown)


client/components/ui-design-system/forms/index.jsx
```diff
@@ -86,6 +86,44 @@ const UIForms = () => (
...
+            <div>{ '<input type="text" placeholder={"Something"} className="input input-is-success" />' }</div>
+          </code>
+        </pre>
+      </div>
+    </div>
+    <div className="row u-mb-0">
+      <div className="columns small-12">
+       <h4>Input Error</h4>
+        <input type="text" placeholder={'Something'} className="input input-has-error"/>
+        <p className="input-error">Contextual alert message goes here</p>
```
marcgurney: Include this in the `pre` code sample underneath too

client/components/ui-design-system/buttons/index.jsx
```diff
@@ -20,25 +20,42 @@ const UIButtons = () => (
...
             <div>{ '<button className="button button-is-danger">Destructive Button</button>' }</div>
             <div>{ '<button className="button" disabled>Disabled Button</button>' }</div>
           </code>
         </pre>
       </div>
     </div>
-
+    <div className="row u-mb-2">
+      <div className="columns small-12">
+        <h3>Dark Background</h3>
```
marcgurney: We should also include a quick sentence in a `p` tag about when this button is used, i.e. when placed on a dark background.

marcgurney:
Two very minor comments, otherwise lgtm!

client/styling/forms.less
```diff
@@ -11,4 +11,38 @@
...
+  &-has-error {
+    border: 1px solid @alertRed;
+    box-shadow: 0 2px 3px rgba(239, 76, 76, 0.2);
+    &:focus {
+    border-color: @alertRed;
+    box-shadow: 0 2px 3px rgba(239, 76, 76, 0.2);
+    }
+  }
+}
+.input-container {
```
marcgurney: This could use a more semantic name, maybe `.input-with-message`, and then `.input-error` should be made an element within this block, similar to `.success-label`.

(sorry, when I reviewed yesterday I forgot that you were making a new component and not just importing from the existing library so I didn't read over the LESS)

marcgurney:
lgtm 👍

PR #9, opened 2017-07-20
js-feat-modal

sampoonachot:
Hey @rawfalafel, I ported the view-modal styles from the design toolkit, but I added some add'l values to "&-close" because it was pulling some styles from another LESS file. I also removed the hr tag and added a border-top and border-bottom on "&-body" instead.


client/styling/view-modal.less
```diff
@@ -0,0 +1,312 @@
...
+
+        &-is-large {
+            font-size: 30px;
+        }
+    }
+
+    // Additional class to win specificity over
+    // i[class^="mms-icon-"]
+    // for font size
+    &-header-icon, &-header i.view-modal-header-icon {
```
rawfalafel: The second rule `&-header i.view-modal-header-icon` exists to override a legacy stylesheet: https://github.com/10gen/mms/blob/master/server/src/webapp-mms/static/less/legacy/core.less#L57-L58

I think we should remove it here since this styleguide is a clean slate.

PR #11, opened 2017-07-21
js-feat-lists

sampoonachot:
FontAwesome and MMS Icon fonts added in LESS file


rawfalafel:
The code LGTM in general, I'll take a look at the build after the fonts are added.

sampoonachot:
@rawfalafel nested-list.less is a bit odd. pulled the code from the design toolkit and .nested-list-link lives outside of .nested-list, but .nested-list is referencing .nested-list-link multiple times in the code.. not sure how to rewrite this if you have any recs?

client/styling/nested-list.less
```diff
@@ -0,0 +1,117 @@
+.nested-list-link {
```
rawfalafel: This code is setup this way in order to reuse the styles. I would make two changes though:
- Change the selector from `.nested-list-link` to `.nested-list-link()` instead. This won't output the style in the compiled css stylesheet: http://lesscss.org/features/#mixins-feature-not-outputting-the-mixin
- Replace [these lines](https://github.com/leafygreen/design/pull/11/files#diff-228ad25f7b328965b8bf8d990ee1475fR14-R18) with `.nested-list-link()`. Better to reuse the mixin everywhere we can.

client/styling/step-indicator.less
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    &-step {
+        font-size: 14px;
+        width: 175px;
+        margin-left: 40px;
+        display: inline;
+        position: relative;
+
+        hr { // specificity battle
```
rawfalafel: I think these are specificity battles with one of the legacy stylesheets in the mms repo. Do we want to keep these here so that these components still work when we use them in mms or do we want to remove them because this is a "clean slate"? @marcgurney

marcgurney: I'd say go with the clean slate version. 

The approach to integrating this design system into the products will be component-by-component. So Cloud would run its "local" `step-indicator` until it's ready to handle the generic one that's part of this system.

sampoonachot:
ready for review

marcgurney:
lgtm 👍

PR #12, opened 2017-07-21
js-feat-rich-radio-w-col

sampoonachot:


client/styling/checklist.less
```diff
@@ -0,0 +1,38 @@
...
+            zoom: 1;
+        }
+    }
+
+    &-item-is-disabled {
```
rawfalafel: [np] This could be nested inside `-item`

sampoonachot: ready for review

PR #13, opened 2017-07-23
Switch to using nwb for all tooling.

hswolff:
This migrates this project to use [nwb](https://github.com/insin/nwb) which will now handle the entire build chain, i.e. webpack and all other tools. This should make it easier to upgrade webpack and other similar tools as nwb handles wrapping and managing all those tools for us, and all we have to do is upgrade our one `nwb` dependency.

I also switched to using hash routing so when you build the app it doesn't require a node server to handle the deep links, as the hash will work with the SPA.

To upgrade to use this you need to

```
rm -rf node_modules
npm install
```

For any first time developers the workflow remains the same.


marcgurney:
Followed instructions to build locally and everything seems to work as expected so lgtm 👍 

Will it be possible to customize URLs later? Perhaps to go from current:
`http://localhost:3000/ui-design-system/forms#/ui-design-system/buttons?_k=95cvov`
to something like:
`http://localhost:3000/ui-design-system/forms#/buttons`

hswolff:
That's from the old URL. Just delete the preceding url path and the hash will work fine. 

Having it this way will now make it so when we `npm run build` that site will still work.

marcgurney:
Niceee

PR #14, opened 2017-07-24
js-feat-table-dark

sampoonachot:
Colin redid the code for tables and didn't use "plain-table.less", so I created the dark theme table based off on his new design


client/styling/tables.less
```diff
@@ -12,15 +12,31 @@ table {
         }
       }
     }
+
```
rawfalafel: This stylesheet should use a class as a selector instead of `table`, we don't want this stylesheet to accidentally mess the styling for a table that can't accept these styles.

sampoonachot: ready for review

client/styling/tables.less
```diff
@@ -1,41 +1,71 @@
...
         border-bottom: none;
       }
     }
+
+    &-is-dark {
+      background-color: @gray1;
+      color: @pw;
+
+      &:hover {
+        td {
```
rawfalafel: this should be replaced with `.table-column` to make it BEM compliant :)

client/styling/tables.less
```diff
@@ -1,41 +1,71 @@
...
+          background-color: @gray4;
+
+          .button {
```
rawfalafel: Do you know what this rule is overriding? I wonder if it's still necessary.

sampoonachot:
ready for review

PR #17, opened 2017-07-26
Updated UI section sidebar navigation to include sections

fredtruman:
screenshot 2017-07-26 11 57 12

@sampoonachot @marcgurney 
Added some sections to the sidebar navigation.

There is a "patterns" section now, where we can add patterns of component use. One that I want to start with is editing an element in Compass. It's a repeatable pattern, but is most established in the Compass CRUD feature. I think I'll spend some time with Matt next week adding the core components (somewhere) and then adding language into the patterns section to illustrate how the component is used.

I also added the "layouts" section where I think we can start hinting at the product-specific app layouts when it makes sense. In our doc, we started thinking about how to break components down by product, but thinking about it some more it probably is a good idea to continue pushing product-agnostic components first and foremost, and then just think about how they might be laid out differently across our products.

Thoughts?

### DIFF
--- a/client/components/ui-design-system/base-styles/colors/index.jsx
+++ b/client/components/ui-design-system/base-styles/colors/index.jsx
@@ -3,7 +3,7 @@
 // ==================================================
 
 import React from 'react';
-import '../../../styling/root.less';
+import '../../../../styling/root.less';
 
 const UIColors = () => (
   <div className="wrap">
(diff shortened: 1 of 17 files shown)


marcgurney:
I agree with the notion of keeping things product-agnostic as much as we can. Maybe we can take more of a tag/filter approach to the IA. So, for example, the Document List (naming?) component would have tags indicating that it's being used in `Compass` `Atlas` `Cloud Manager`. Could start with just the tags/labels and in future versions of the site users could filter to "only show me components used in Atlas".

LGTM 👍 for this PR.

PR #18, opened 2017-07-27
js-feat-home

sampoonachot:


client/components/ui-design-system/welcome/index.jsx
```diff
@@ -7,8 +7,65 @@ import '../../../styling/root.less';
...
+				  	<h1 className="home-header">Some copy welcoming someone to the site.</h1>
+				  	<a className="link home-link">Explore the Design System</a>
+			  	</div>
+		  	</div>
+	  	</div>
+  	</div>
+
+  	<div className="row u-mb-9">
+	  	<div className="columns small-7">
+		  	<img className="home-img"></img>
```
marcgurney: `` tags are self-closing so these should be written like:
```
<img className="home-img" />
```

PR #19, opened 2017-08-17
js-feat-obscurable-input

sampoonachot:
@rawfalafel is there a way to add a class name to calc() in css? see `forms.less` 

Instead of using 90px I want to use the length of the `.input-message`

```
  &-has-form-control {
    padding-right: 90px;
  }
```


rawfalafel:
You would have to give `input-with-message` a fixed length and use that.

PR #22, opened 2017-08-22
js-feat-navigation

sampoonachot:


client/styling/navigation.less
```diff
@@ -0,0 +1,48 @@
+.navigation {
+  float: right;
+
+  a {
```
rawfalafel: I was hoping this would be an actual BEM component but this is a start :)

sampoonachot: oh god is it not. ok hold up

sampoonachot: @rawfalafel rdy

rawfalafel:
there's a merge conflict 😭

client/app.jsx
```diff
@@ -2,11 +2,11 @@ import React, { Component } from 'react'
 import Root from './components/root.jsx'
 
 class App extends Component {
-  render() {
-    return (
-      <Root />
-    );
-  }
+ render() {
```
rawfalafel: Looks like this diff only has one line?

PR #28, opened 2017-08-29
js-feat-layouts

sampoonachot:
still figuring how to install that font awesome bit


client/styling/empty-view.less
```diff
@@ -0,0 +1,143 @@
...
+  padding: 50px calc(~"50% - 400px") 50px;
+
+  &-margin {
+    margin: 35px 0;
+  }
+
+  &-image {
+    padding: 12px 0 10px;
+
+    img {
```
rawfalafel: This doesn't look like it's used anywhere

sampoonachot: removed! should i add font awesome in a diff branch

rawfalafel: same branch, the diff should be small

PR #29, opened 2017-08-30
PD-10: Guidelines/patterns for modifying elements

fredtruman:
- I'm not sure about the correct markup to use to get the right visual hierarchy, but I didn't touch any of that CSS.
- This is very functional language
- Gifs are quick. Open to converging on a style if we think this is something we want to replicate in other guides
- GIFs are also large file size. Not 100% on using them, but they seem to work better than just lots of images inline with text. The content gets lost. Maybe we try a two column approach with content on left and images on right or something.

**SCREENSHOTS:**

screenshot 2017-08-30 16 48 15
screenshot 2017-08-30 16 48 23
screenshot 2017-08-30 16 48 31
screenshot 2017-08-30 16 48 40
screenshot 2017-08-30 16 48 50


marcgurney:
This is awesome. 

I'd personally prefer breaking out the interaction stages into a storyboard grid. I think that would make it easier to study.

fredtruman:
You might be right 🤔  will think on how to do this

sampoonachot:
Love this! Great work, Fred!

This might just be me, but I think I had to look at the gif, read the "Description of pattern", then look at the "When you should use..." in that order in order to understand the pattern usage. 

Also the way this is laid out seems like these modifying patterns are all connected but they're all their own individual patterns. so maybe instead of using the generic name of Modification Pattern 1, we can title them as what they are e.g. Modifying title (for MongoDB Charts). Also might be good to have a way to distinguish/filter between MongoDB Cloud, Charts, Stitch, University guidelines.

fredtruman:
> This might just be me, but I think I had to look at the gif, read the "Description of pattern", then look at the "When you should use..." in that order in order to understand the pattern usage.

Ah ok. I can think about this more in the layout 🤔 

> Also the way this is laid out seems like these modifying patterns are all connected but they're all their own individual patterns. so maybe instead of using the generic name of Modification Pattern 1, we can title them as what they are e.g. Modifying title (for MongoDB Charts). Also might be good to have a way to distinguish/filter between MongoDB Cloud, Charts, Stitch, University guidelines.

Well... they're not unconnected though. **What this "Guidelines" section is:** when a designer is tasked with designing some mechanism to modify an element, they can look at this guide and depending on the element's criteria, they can decide which pattern is appropriate. 

**What it's not really intending to do:** be specific and comprehensive documentation for the instances of these patterns in our different apps. The apps themselves should be documentation enough for this (one or two text examples is helpful I think though).

The guidelines for modification patterns should be cross-product and the patterns should be a bit open-ended I think. I'm thinking of these as Class definitions vs. Instantiated objects. We can discuss this more in DS sync or IRL 😄

fredtruman:
@marcgurney @sampoonachot RFAL 👀 

- I moved the "when should you use" section beneath the description of the pattern, as per Joy's suggestion.
- embedded static images inline with the pattern description rather than animated gifs on their own
- a couple global typography spacing tweaks
- a few more lines of descriptive language in each section

PR #30, opened 2017-09-05
js-hotfix-nested-list

sampoonachot:
there was some styling in `root.less` for `ul` and `li` that was applying to `nested-list.less` so I removed that.


client/styling/root.less
```diff
@@ -64,13 +64,6 @@ textarea {
   resize: none;
 }
 
-ul, li {
```
rawfalafel: 👍 Do you know where the other rules above come from? We probably want to remove those too, if they're interfering with other BEM styles

client/styling/nested-list.less
```diff
@@ -10,83 +10,26 @@
...
     .nested-list-link();
     padding: 8px;
 
-    &::before {
-      content: "\E308";
-      font-family: "MMS Icons";
-      color: @gray5;
-      margin-right: 5px;
-    }
+    // &::before {
```
rawfalafel: Did you mean to delete this?

client/components/ui-design-system/index.jsx
```diff
@@ -10,36 +10,35 @@ const UIDesignSystem = ({ children }) => (
   <nav className="navigation-sidebar">
     <ul className="navigation navigation-is-vertical navigation-is-vertical-parent">
       <li><h4>Product Design</h4>
-       <ul className="navigation-is-vertical-parent">
-        <li><Link to='/ui-design-system/welcome' className="navigation-link">Welcome</Link></li>
-       </ul>
+        <ul className="navigation-is-vertical-parent">
```
rawfalafel: the indentation looks weird 👀

client/styling/nested-list.less
sampoonachot: yeah wanted to keep it in until i merged the mms-icons

client/components/ui-design-system/index.jsx
sampoonachot: i fixed it in the recent commit

client/styling/root.less
sampoonachot: `pre` is being used for the code snippets.. `img` in multiple files?

rawfalafel: I think the styles in `input` might be overwriting the styles [here](https://github.com/leafygreen/design/blob/master/client/styling/forms.less#L2)

sampoonachot: 👍

PR #33, opened 2017-09-07
tab section update

ingridh:
updated the tabs section with three 'common navs'


client/components/ui-design-system/components/tabs/index.jsx
```diff
@@ -58,6 +58,256 @@ const UITabs = () => (
...
+    <div className="row u-mb-2">
+      <div className="columns small-12">
+        <h3>Common Navigation</h3>
+        <p>Section Header, Tabs, Subtabs, Controls and Banner</p>
+      </div>
+    </div>
+
+    <div className="row">
+      <div className="sectionBox">
+        <div className="sectionTitle"><h2>Section With Tabs</h2>
```
rawfalafel: The elements here should use the same CSS classes that are used in the `<code>` block below. So I believe this section should copy and paste the HTML in there.

ingridh:
Updated with BEM styling. Let me know if there's anything else that need to be fixed!

client/styling/tabs.less
```diff
@@ -31,5 +31,101 @@
...
+    }
+
+    &-more {
+      font-size: 12px;
+      top: 7px;
+    }
+  }
+}
+
+.section-header {
```
rawfalafel: The rule is to create a single stylesheet for each BEM component. This file is for the `tabs` component, and we should create a new file for `section-header` and `section-banner`. You'll notice that we do that in the mms repo [here](https://github.com/10gen/mms/blob/master/client/less/bem-components/section-banner.less) and [here](https://github.com/10gen/mms/blob/master/client/less/bem-components/section-header.less).

client/styling/tabs.less
```diff
@@ -31,5 +31,101 @@
...
+  width: 100%;
+  padding: 1.3rem;
+  margin: 0 0.5rem;
+
+  &-title {
+    display: flow-root;
+    margin-bottom: 1rem;
+  }
+
+  h2 {
```
rawfalafel: One of the goals of BEM is to avoid CSS rules that could have unintended consequences. We only allow CSS selectors that start with the name of the component (in this case `.section-header`) to avoid unintentionally styling elements. For example, we might want to add another `h2` element inside `section-header` that's styled differently. That wouldn't be possible here because this rule encompasses all `h2` elements. We avoid that by, for example, selecting this with `&-subheader`. 

If you look at the existing [section-header.less](https://github.com/10gen/mms/blob/master/client/less/bem-components/section-header.less) stylesheet, you'll notice that all of the rules (with a few commented exceptions) follow the same `&-somename` pattern. We should continue that here too.

ingridh:
ok, I've created three other .less files (section-subtabs, section-controls, section-header, section-banner) for the appropriate styling. I copied a few things from the mms sheets but not everything because not everything was needed

thoughts? 🤔

client/styling/section-header.less
```diff
@@ -0,0 +1,30 @@
...
+    margin-bottom: 1rem;
+
+    &-text {
+      float: left;
+      font-size: 1.625rem;
+      font-weight: bold;
+    }
+  }
+
+  .alert {
```
rawfalafel: There's an `alert` BEM component in `mms`. We should bring in that stylesheet if we're going to use the alert component here

client/styling/section-controls.less
```diff
@@ -0,0 +1,20 @@
+.section-controls {
+  margin: 0.7rem 0;
+
+  &-box{
+    display: inline !important;
```
rawfalafel: We should avoid `!important`. What style is this conflicting with that's forcing you to use it?

client/styling/section-banner.less
```diff
@@ -0,0 +1,19 @@
...
+    &-keyword {
+        color: @green2;
+        font-size: 13px;
+        text-transform: none;
+    }
+
+    &-item {
+        margin-right: 20px;
+    }
+}
```
rawfalafel: There should be a way for your editor to automatically add a newline at the end of the file. For example, Atom has a package called [whitespace](https://atom.io/packages/whitespace)

client/styling/section-header.less
```diff
@@ -0,0 +1,30 @@
+.section-header {
```
marcgurney: I took a look at `section-header.less` in the Cloud codebase and it looks different enough that I think it might be best to just remove the component less files you added in this PR and replace with the versions from Cloud.

https://github.com/10gen/mms/blob/master/client/less/bem-components/section-header.less

If the markup lines up 1:1 then it should render correctly.

PR #37, opened 2017-09-08
Layout Page Bug Fixes

Sean-Oh:
This PR fixes some minor bugs that were introduced in a past [PR](https://github.com/leafygreen/design/pull/28) that broke the Layout page.

I also added some of the missing section component `.less` files. However, some of the basic components (inputs and dropdowns) still aren't rendering properly. These are probably React components, but we're just showing the rendered HTML in the design system page. We'll probably want to replace the rendered HTML with the actual React component. 

screen shot 2017-09-08 at 4 04 54 pm


Sean-Oh:
Merging for now. We need to uncomment `Section with Tabs` in the future.

PR #47, opened 2017-10-25
Content updates

marcgurney:
* Removed brand guide, link to blog
* Linked to Brand Portal in Resources section
* Updated copy on Home and Welcome pages
* Updated copy in footer

Not final final, but closer to something that could be made public.


Sean-Oh:
Ran into compile issues because the related brand resource links weren't removed from the `routes.jsx`. Removing them now. 
screen shot 2017-10-25 at 5 54 56 pm

marcgurney:
Ah, thank you.

PR #50, opened 2017-10-27
Updated green

meltothewong:
Add darker greens and tones. @marcgurney Does this look good to you?

### DIFF
--- a/client/components/ui-design-system/base-styles/colors/index.jsx
+++ b/client/components/ui-design-system/base-styles/colors/index.jsx
@@ -66,7 +66,7 @@ const UIColors = () => (
         <div className="color-chip">
           <span className="color-chip-swatch green8"></span>
           <span className="color-chip-code code">@green8</span>
-          <p className="color-chip-hex">Hex: #F0F7EC</p>
+          <p className="color-chip-hex">Hex: #EFF6EC</p>
         </div>
       </div>
     </div>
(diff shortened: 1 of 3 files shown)


client/components/ui-design-system/base-styles/colors/index.jsx
```diff
@@ -52,21 +52,21 @@ const UIColors = () => (
...
-          <p className="color-chip-hex">Hex: #c4dbb3</p>
+          <p className="color-chip-hex">Hex: #C2E0B4</p>
         </div>
       </div>
       <div className="columns small-3">
         <div className="color-chip">
           <span className="color-chip-swatch green8"></span>
           <span className="color-chip-code code">@green8</span>
-          <p className="color-chip-hex">Hex: #f0f7ec</p>
+          <p className="color-chip-hex">Hex: #F0F7EC</p>
```
Sean-Oh: This @green8 is the old hex that you deleted - should it be #EFF6EC instead?

meltothewong:
@Sean-Oh Hmm I don't remember making that change, but yes, that's correct. Thanks for catching that!

client/components/ui-design-system/base-styles/colors/index.jsx
meltothewong: @Sean-Oh Hmm I don't remember making that change, but yes, that's correct. Thanks for catching that!

marcgurney:
LGTM

PR #53, opened 2017-10-30
Copy pass, added links to sketch templates

marcgurney:


marcgurney:
btw the change to `package.json` is actually no change, the `--save` flag on an `npm install` just reordered some items.

PR #57, opened 2017-10-31
Added react component for layouts

sampoonachot:


sampoonachot:
@Sean-Oh not exactly sure why `this.graphic = (<div className="graphic-test" />);` is needed in react-layouts.jsx

Sean-Oh:
@joy You can just `return null` instead of returning an empty `div`, since this will get passed as a node to the fallback prop of the `SpashViewGraphic` component. 

This LGTM. Other changes I added were:
- Adjusted the button sizes to match the CSS page.
- Remove {this.children} and {this.graphic}. The user wouldn't know what these are when looking at the code snippet. It's okay to just add content directly inside react components.

PR #61, opened 2017-10-31
sanity checked html and react for modal component

sampoonachot:
- replaced `div` with `header` and `footer` to match react component
- removed border padding on html component and added hr tags


Sean-Oh:
LGTM

PR #68, opened 2017-11-01
React lists

chrissygonzalez:
I got the paginated table to show up! Should probably update the dummy data to match the css version.

Still having trouble passing the exact same styles through to the table cells, but added the standard table to the Tables section as well.


Sean-Oh:
LGTM - added the changes we talked about. Great job @chrissygonzalez! ✨ ✨

PR #77, opened 2017-11-06
Mms icons in mms-fonts branch

chrissygonzalez:
Still need to add the marketing icons and maybe link out to Material and Font Awesome.


marcgurney:
I pushed a small change to remove the unused "stacked" icon and the colors it relied on.

It would be nice to give this table some extra styling to make it look more like the table in /admin (since this is a special-purpose table).

PR #79, opened 2017-11-07
File organization

marcgurney:
screen shot 2017-11-08 at 12 08 10 pm


marcgurney:
Fixed the tab alignment in this PR too.

marcgurney:
Good call. Renamed `bem-components` to `css`, divided that directory into `global`, `bem-components`, and `site`.

PR #86, opened 2017-12-13
PD-14: Button Size Adjustments

fredtruman:
@marcgurney || @sampoonachot 

Adjustments to:
* XS button size - NO CHANGE
* SM button size - 27px height; font-size 11px; ALL CAPS
* Normal button size - change from 37px height to 32px height
* LG button size - decrease from 47px height to 45px height

**BEFORE**
screenshot 2017-12-13 09 00 52

**AFTER**
screenshot 2017-12-13 09 01 06

cc @meltothewong

### DIFF
--- a/client/css/bem-components/buttons.less
+++ b/client/css/bem-components/buttons.less
@@ -8,7 +8,7 @@
   display: inline-block;
   font-size: 14px;
   font-weight: normal;
-  padding: 9px 12px;
+  padding: 6px 12px 7px 12px;
   text-decoration: none;
   cursor: pointer;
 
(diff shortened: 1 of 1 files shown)


marcgurney:
The small and normal buttons seem a little short. In the "before" there is a more gradual increase in height from xs -> large. In the "after" the xs, small, and normal are pretty close in height, and then it jumps for the large. Thoughts on adding just a few more px top/bottom padding to small and normal?

fredtruman:
I sort of think 32px is the right Normal size. 🤔  
Maybe this -lg is really an -xl and there is an as-of-yet-defined -lg size? 😅

marcgurney:
How about 28px and 34px for SM and Normal?

fredtruman:
The math is sound. I can try

fredtruman:
screenshot 2017-12-13 10 55 02

This is 28px for small and 34px for Normal. It's ok. The downsides are:
- Small is slight top heavy (or alternatively bottom-heavy) with 9px of padding on top and 8px of padding on bottom (or vice versa). Since the font-size is 11px, having the odd-height of 27px allows for an even split of 8px on top and bottom of the type.
- Normal is already 32px basically everywhere is production.

marcgurney:
Fair enough, I'm fine with what you had in your initial comment. lgtm 👍

PR #92, opened 2018-05-04
Updated _university-template-1.2 to 1.5

meltothewong:


meltothewong:
@marcgurney Noticed my template was waaaaay behind. Found out it was pulling the wrong Sketch file name.

PR #94, opened 2018-07-27
PD-19: First pass at adding new zero states

fredtruman:
screenshot 2018-07-27 14 00 20
screenshot 2018-07-27 14 00 28


ingridh:
I feel like the XL? buttons look really big.. thoughts? Otherwise, it looks really good 👏🏼

PR #97, opened 2018-08-02
PD-38 updated copy snippet

sampoonachot:
- Updated copy snippet for both single and multi-line codes.
- Updated inset box-shadow of button

screen shot 2018-08-03 at 1 57 14 pm
screen shot 2018-08-03 at 1 57 26 pm
screen shot 2018-08-03 at 1 57 33 pm


sampoonachot:
- [x] looks like too much drop shadow on the button? `I removed the dropshadow`
- [x] shouldn't our buttons have a thicker bottom border too? `added box-shadow inset to our buttons`
- [x] should we make the height of the code "banner" match the height of our other alert banners? I know they're different, but something to consider `yup! height of single snippet is now 54px`
- [x] I wonder if we can apply all the positioning styles to the button as flexbox rather than absolute positioning. I imagine it would be more compatible in more situations. `code snippet now uses flexbox`
- [ ] Can we come up with any size variations that could work for similar application like: https://mongodb.invisionapp.com/share/5YLO0RI8FAX#/305176194_50_03 is this a good idea? `i think this is a separate component to explore`

PR #106, opened 2019-08-29
PD-353: Add banner indicating content is out-of-date

DesignerDave:
With the upcoming revamp of the site, this banner is a temporary indicator to reduce internal confusion about what resources are correct to use.

Here's what the banners look like:
Screen Shot 2019-08-29 at 4 04 06 PM

The links are only visible if there is already a resource to replace the existing content.


DesignerDave:
Chatted in person –
- Yes
- No – this is the safe implementation.

PR #131, opened 2022-08-22
Handles content loading errors

TheSonOfThomp:


.env.example
```diff
@@ -0,0 +1,4 @@
+CONTENTFUL_SPACE_ID='s9zggw79q9w3'
+CONTENTFUL_ACCESS_TOKEN='<Personal access token>'
+CONTENTFUL_DELIVERY_TOKEN='<Content Delivery API - access token>'
+CONTENTFUL_PREVIEW_TOKEN='<Content Preview API - access token>'
```
spark33, line 4: I like this!

utils/getContentfulResources.ts
```diff
@@ -2,13 +2,17 @@ import { ContentfulClientApi, EntryCollection } from 'contentful';
...
 const contentful = require('contentful');
+const isProd = process.env.NODE_ENV === 'production'
 
 export function createContentfulClient(): ContentfulClientApi {
-  return contentful.createClient({
-    space: '5fb8fg2odr3q',
+  const client = contentful.createClient({
     environment: 'master',
-    accessToken: process.env.CONTENTFUL_ACCESS_TOKEN as string,
+    space: process.env.CONTENTFUL_SPACE_ID ?? 's9zggw79q9w3',
```
spark33, line 10: I don't like having this be a string value here, even if it's just a backup value. The request would fail without the other env variables being set properly anyway.

utils/getContentfulResources.ts
```diff
@@ -45,15 +52,18 @@ export async function getComponent(componentKebabCaseName: string) {
...
-    throw error;
+    console.error('No Page Groups found', error)
+    // Return no sections
+    return [] as EntryCollection<ContentPageSectionFields>['items']
   }
 }
 
 export async function getContentPage(contentPageGroupTitle: string, contentPageTitle: string) {
+  // eslint-disable-next-line no-console
   console.log(contentPageGroupTitle, contentPageTitle)
```
spark33, line 67: We should just remove this now

spark33 (changes requested):
Some minor comments - sorry I accidentally didn't tie them to the review.

utils/getContentfulResources.ts
```diff
@@ -2,13 +2,17 @@ import { ContentfulClientApi, EntryCollection } from 'contentful';
...
 
 export function createContentfulClient(): ContentfulClientApi {
-  return contentful.createClient({
-    space: '5fb8fg2odr3q',
+  const client = contentful.createClient({
     environment: 'master',
-    accessToken: process.env.CONTENTFUL_ACCESS_TOKEN as string,
+    space: process.env.CONTENTFUL_SPACE_ID ?? 's9zggw79q9w3',
+    accessToken: isProd ? process.env.CONTENTFUL_DELIVERY_TOKEN : process.env.CONTENTFUL_PREVIEW_TOKEN,
+    host: isProd ? 'cdn.contentful.com' : 'preview.contentful.com',
```
spark33, line 12: I think it would be nicer to have a object that looks like `{ production: { host: '...', ...}, staging: {...}`. Would scale better to potentially actually having preview or testing environments.

utils/getContentfulResources.ts
TheSonOfThomp: True

PR #139, opened 2022-08-25
PD-2350 TSDoc prop tables

TheSonOfThomp:


pages/component/[componentName]/documentation.tsx
```diff
@@ -31,11 +35,38 @@ export async function getStaticProps({ params }) {
...
+        // And that have prop definitions
+        .filter(doc => Object.keys(doc.props).length > 0)
+        // and are not tagged as internal
+        .filter(doc => isUndefined(doc.tags?.internal)),
+      'displayName',
+    );
+  } catch (error) {
+    tsDoc = null;
+  }
+
```
Comment (spark33) [inline pages/component/[componentName]/documentation.tsx:63]: I feel like this might fit well in `getDependencyDocumentation`, since that's already a function that reads files from within the package directories - thoughts?

Comment (TheSonOfThomp) [reply pages/component/[componentName]/documentation.tsx:58]: Yeah that would make sense 👍

PR #148, opened 2022-09-09
TSDoc updates

TheSonOfThomp:


components/pages/documentation/TSDocPropTable/TSDocPropTable.tsx
```diff
@@ -0,0 +1,267 @@
...
+  if (groupName === 'HTMLAttributes')
+    return 'https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes';
+
+  let tag = groupName
+    .slice(0, groupName.indexOf('HTMLAttributes'))
+    .toLowerCase();
+
+  tag = tag == 'anchor' ? 'a' : tag;
+  return `https://developer.mozilla.org/en-US/docs/Web/HTML/Element/${tag}`;
+}
```
spark33, line 267: Could we break out the `css` definitions and different functions/components to separate files?

spark33 (changes requested):
Small change request to break out one of the large files!

components/pages/documentation/TSDocPropTable/TSDocPropsTable.types.ts
```diff
@@ -0,0 +1,37 @@
...
+  props: PropCategory;
+};
+
+export interface PropGroup {
+  groupName: string;
+  props: Array<PropItem>;
+}
+
+export const isPropItem = (obj: any): obj is PropItem => {
+  return (
```
DesignerDave, line 30: Nit – since all we do is return this boolean, this function can be simplified to:

```js
export const isPropItem = (obj: any): obj is PropItem => (
  !isUndefined(obj.name) &&
  !isUndefined(obj.required) &&
  !isUndefined(obj.type) &&
  !isUndefined(obj.description) &&
  !isUndefined(obj.defaultValue)
);
```

We can also perhaps take it a step further:
```js
export const isPropItem = (obj: any): obj is PropItem => [
  obj.name,
  obj.required,
  obj.type,
  obj.description,
  obj.defaultValue,
].every(property => !isUndefined(property));
```

DesignerDave (commented):
Had some notes – mostly code cleanup

components/pages/documentation/TSDocPropTable/PropTableSection.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const TSDocPropTableSection = ({
+  tsDocArray,
+}: {
+  tsDocArray: Array<CustomComponentDoc>;
+}) => (
+  <>
+    {tsDocArray?.map(doc => (
+      <TSDocPropTable
+        key={doc.displayName}
```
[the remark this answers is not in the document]
TheSonOfThomp: They're by definition unique in this case, I believe. The display name is the original key used to generate the array

components/pages/documentation/TSDocPropTable/PropTableTooltipContent.tsx
```diff
@@ -0,0 +1,114 @@
...
+    </InlineCode>
+
+    <div>
+      <Label htmlFor={`${prop.name}-type`} darkMode>
+        Type: &nbsp;
+      </Label>
+      <InlineCode
+        id={`${prop.name}-type`}
+        className={css`
+          // This should be default
```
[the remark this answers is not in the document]
TheSonOfThomp: Not yet. This is the ticket: https://jira.mongodb.org/browse/PD-2424

components/pages/documentation/TSDocPropTable/PropTableTooltipContent.tsx
```diff
@@ -0,0 +1,114 @@
...
+  font-size: ${typeScales.code2.fontSize}px;
+  display: inline-block;
+`;
+
+const dividerStyle = css`
+  border-color: ${palette.gray.dark1};
+  margin: ${spacing[2]}px -${spacing[3]}px;
+`;
+
+export const PropTableTooltipContent = ({ prop }: { prop: PropItem }) => (
```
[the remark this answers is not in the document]
TheSonOfThomp: Ha, that's fair

components/pages/documentation/TSDocPropTable/TSDocPropTable.tsx
```diff
@@ -0,0 +1,201 @@
...
+  );
+
+  const props = [...componentProps, ...inheritedProps];
+
+  return (
+    <>
+      <ExpandableCard
+        title={`${tsDoc?.displayName} props`}
+        /// @ts-ignore
+        // description={<Markdown>{tsDoc?.description}</Markdown>}
```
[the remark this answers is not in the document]
TheSonOfThomp: the `ts-ignore` is because currently `description` currently does not accept `ReactNode`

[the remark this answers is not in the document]
TheSonOfThomp: We currently have a lot of messy `description` text in TSDoc, so I'm going to just remove this for now

components/pages/documentation/TSDocPropTable/TSDocPropsTable.types.ts
TheSonOfThomp: I like that array method 👍

components/pages/documentation/TSDocPropTable/PropTableSection.tsx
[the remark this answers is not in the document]
DesignerDave: Sounds good

components/pages/documentation/TSDocPropTable/PropTableTooltipContent.tsx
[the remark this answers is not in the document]
DesignerDave: Ahh okay, thanks!

components/pages/documentation/TSDocPropTable/PropTableTooltipContent.tsx
```diff
@@ -0,0 +1,92 @@
...
+  </div>
+);
+
+/**
+ * Temporarily duplicating this function
+ *
+ * TODO: Move this to TSDoc utils
+ */
+
+function getDefaultValueString(defaultValue: any): string {
```
DesignerDave, line 82: Flagging this type since it's `any` and not a typechecking function. Why does this type need to be so broad?

components/pages/documentation/TSDocPropTable/TSDocPropTable.tsx
```diff
@@ -0,0 +1,169 @@
...
+const requiredHighlightStyle = css`
+  padding-left: 1ch;
+  color: ${palette.red.base};
+  text-transform: uppercase;
+
+  &::before {
+    content: '(';
+  }
+  &::after {
+    content: ')';
```
DesignerDave, line 34: Using pseudo-elements here feels a little clever. We should probably just use a STL to add these.

pages/component/[componentName]/documentation.tsx
```diff
@@ -4,42 +4,64 @@ import { getDependencyDocumentation } from 'utils/_getComponentResources';
...
+export async function getStaticProps({ params: { componentName } }) {
   const {
     props: { changelog, readme, tsDoc },
-  } = await getDependencyDocumentation(params.componentName);
+  } = await getDependencyDocumentation(componentName);
+
+  // Here we pull out the designGuidelines to ensure we're not passing that to this page unnecessarily
+  const {
+    fields: { designGuidelines, ...meta },
+  } = (await getComponent(componentName)) || {
```
Comment (DesignerDave) [inline pages/component/[componentName]/documentation.tsx:54]: Nit: _Technically_ we should probably use nullish coalescing (`??`) rather than pipes here I think.

components/pages/documentation/TSDocPropTable/TSDocPropTable.tsx
```diff
@@ -0,0 +1,169 @@
...
+  tsDoc,
+  className,
+}: {
+  tsDoc: CustomComponentDoc;
+  className?: string;
+}) => {
+  const componentProps = Object.values(omitBy(tsDoc.props, isInheritableGroup))
+    .flatMap((prop: Props) => Object.values(prop))
+    .sort((a, z) => {
+      if (isRequired(a) && !isRequired(z)) return -1;
```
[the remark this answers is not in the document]
TheSonOfThomp: No, since in the second check we know `a` is not required. If both are required we want to sort in alpha order

components/pages/documentation/TSDocPropTable/PropTableTooltipContent.tsx
TheSonOfThomp: Technically should be `PropItem['defaultValue']`, but that itself is typed `any`

pages/component/[componentName]/documentation.tsx
Comment (TheSonOfThomp) [reply pages/component/[componentName]/documentation.tsx:54]: You're right 👍

components/pages/documentation/TSDocPropTable/TSDocPropTable.tsx
TheSonOfThomp: Not sure I agree pseudo-elements here is overly clever. This feels like a case they're designed for

components/pages/documentation/TSDocPropTable/TSDocPropTable.tsx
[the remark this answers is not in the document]
DesignerDave: You're right 🤦‍♂️ nevermind!

components/pages/documentation/TSDocPropTable/TSDocPropTable.tsx
DesignerDave: I mean technically yes, they are built to be able to do this. Pseudo-elements are also less accessible, and arguably more complex when you're reading the code – for someone looking at the code for the first time, it is certainly harder to know where the parentheses are coming from.

spark33: I like this approach for when the main text is weirdly positioned or difficult to access because it's dynamically rendered, but that doesn't seem to be the case here. Why would we choose this pseudo-element approach over just adding it to our markup?

spark33 (approved):
Generally looks good to me! One minor comment on the pseudo-elements on required props.

components/pages/documentation/TSDocPropTable/TSDocPropTable.tsx
TheSonOfThomp: I used pseudo-elements here to keep the formatting in one place. I wasn't sure whether we'd want the parens and/or all-caps, so I used CSS to do this transformation. 

If we're good with the format `(REQUIRED)` I can hard-code that in the markup

spark33: I'm a little confused. Does text-transform changing the formatting of parens? I thought it would only affect characters in `a-z`. If that's the case, we could move the `()` and keep the rest of the css block as is.

TheSonOfThomp: Also I personally find this hard to read with the parents in the markup:
```jsx
<sup className={requiredHighlightStyle}>(REQUIRED)</sup>
```

DesignerDave: I vastly prefer this TBH. Parentheses are content, not formatting.

PR #150, opened 2022-09-09
PD-2385 Live example

TheSonOfThomp:


spark33:
Could we also rename the `example` directory back to `live-example`?

TheSonOfThomp:
> Could we also rename the `example` directory back to `live-example`?

I named it `example` to match the page name. Why do you recommend changing it back?

spark33:
To me, `example` has a different meaning from `live-example`. `live-example` feels like a unique entity as a feature we offer, while `example` feels just like a word, easily replaced by something `sample`. I also associate it with something that's just there for reference, like `env.example`

TheSonOfThomp:
> To me, `example` has a different meaning from `live-example`. `live-example` feels like a unique entity as a feature we offer, while `example` feels just like a word, easily replaced by something `sample`. I also associate it with something that's just there for reference, like `env.example`

In principle I agree, but I feel since we have `pages/[componentName]/example.tsx` we should also have the `components/pages/example/*` directory. 

Also, to be clear, I don't think we should change the file name in `pages/` since that maps to the URL

utils/getComponentStory.ts
```diff
@@ -0,0 +1,20 @@
...
+} & {
+  [key: string]: ComponentStoryFn<any>;
+};
+
+export async function getComponentStory(
+  kebabName: string,
+): Promise<ModuleType | undefined> {
+  try {
+    return import(
+      `@leafygreen-ui/${kebabName}/src/${pascalcase(kebabName)}.story`
```
spark33, line 15: There may be some files ending in `stories.tsx`, and there's a case to have `stories` be the standard actually - it's what's used by Storybook

utils/tsdoc.utils.ts
```diff
@@ -1,5 +1,63 @@
...
+};
+
+export const getInheritedProps = (props: PropCategories): Array<PropGroup> => {
+  return Object.entries(
+    pickBy(props, isInheritableGroup),
+  ).map(([groupName, props]: [string, Props]) => ({
+    groupName,
+    props: Object.values(props).flatMap(prop => prop),
+  }));
+};
```
spark33, line 60: 🚀🚀🚀

components/pages/example/KnobRow.tsx
```diff
@@ -0,0 +1,86 @@
...
+        &:not(:last-of-type) {
+          border-bottom: 1px solid
+            ${darkMode ? palette.gray.dark2 : palette.gray.light2};
+        }
+
+        & * p {
+          margin: unset;
+        }
+      `}
+    >
```
spark33, line 52: There were changes in recent PRs that makes this flexbox mobile responsive - the label and input should stack in mobile. Could we add this in?

components/pages/example/KnobRow.tsx
```diff
@@ -0,0 +1,86 @@
...
+            spacing={spacing[4]}
+            darkMode={darkMode}
+            definition={
+              <Markdown darkMode={!darkMode}>
+                {argType?.description || prop.description}
+              </Markdown>
+            }
+          >
+            {prop.name}
+          </InlineDefinition>
```
spark33, line 66: Love this!

components/pages/example/Knob.tsx
```diff
@@ -0,0 +1,154 @@
...
+      return (
+        /// @ts-expect-error
+        <Toggle {...rest} checked={!!value as boolean} onChange={onChange} />
+      );
+
+    case 'array':
+    case 'enum':
+    case 'select':
+    case 'radio': {
+      const argOptions = argType?.options
```
spark33, line 78: If arguments from the story are being read in like this, I'm wondering if we can rename `argTypes` inside Knob. I would expect this object we're reading from to be something like `{ ...story.argTypes, ...story.args }`, and while I trust that this is being done elsewhere in the codebase, naming it `argTypes` here doesn't make this clear.

components/pages/example/LiveExample.tsx
```diff
@@ -0,0 +1,140 @@
...
+    const kebabName = kebabCase(componentName);
+    getComponentStory(kebabName)
+      .then(module => {
+        if (module) {
+          const { default: meta, ...stories } = module;
+          const defaultStoryName = meta?.parameters?.default;
+          const StoryFn = defaultStoryName
+            ? stories[defaultStoryName]
+            : Object.values(stories)[0];
+          const args = { ...meta.args, ...StoryFn?.args };
```
spark33, line 70: Yep, here's the overwriting logic - nice. So just renaming the prop in `Knob` would be 👌

components/pages/example/LiveExample.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+          // console.log({ meta, args, StoryFn });
```
spark33, line 72: Unremoved comment!

components/pages/example/LiveExample.tsx
```diff
@@ -0,0 +1,140 @@
...
+        ...state,
+        ...newState,
+      };
+    },
+    {
+      meta: undefined,
+      args: undefined,
+      StoryFn: undefined,
+    } as LiveExampleState,
+  );
```
spark33, line 57: I don't know if you got this from the React docs or whatever, but this is really smart!

components/pages/example/LiveExample.tsx
```diff
@@ -0,0 +1,140 @@
...
+      `}
+    >
+      <div
+        className={css`
+          display: flex;
+          align-items: center;
+          justify-content: center;
+          min-height: 33vh;
+        `}
+      >
```
spark33, line 113: Do we not need more logic on this container to maintain things like `z-index` and preventing overflow from whatever is in the StoryFn? Without having tested the rendering of all components, I'm wondering if we need to move over the global CSS from our Storybook config.

components/pages/example/LiveExample.tsx
```diff
@@ -0,0 +1,140 @@
...
+          knobProps.map(prop => (
+            <KnobRow
+              key={prop.name}
+              prop={prop}
+              darkMode={darkMode}
+              argType={{
+                ...meta?.argTypes?.[prop.name],
+                ...StoryFn?.argTypes?.[prop.name],
+              }}
+              args={args}
```
spark33, line 127: I find these lines a little confusing - didn't we fold the `argType` fields into `args`? Why do we need to pass `argTypes` separately? Doesn't mean that when we consume `argTypes` in `Knob`, we're not getting the fields in `args`?

spark33 (changes requested):

package.json
```diff
@@ -24,63 +25,64 @@
...
+    "@leafygreen-ui/stepper": "^3.1.0-next.2",
+    "@leafygreen-ui/table": "^6.1.0-next.2",
+    "@leafygreen-ui/tabs": "^9.1.0-next.2",
+    "@leafygreen-ui/text-area": "^6.2.0-next.2",
+    "@leafygreen-ui/text-input": "^10.2.0-next.2",
+    "@leafygreen-ui/toast": "^2.0.1-next.3",
+    "@leafygreen-ui/toggle": "^8.0.4-next.2",
+    "@leafygreen-ui/tokens": "^1.3.3-next.2",
+    "@leafygreen-ui/tooltip": "^7.1.2-next.2",
+    "@leafygreen-ui/typography": "^13.2.0-next.2",
```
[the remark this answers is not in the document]
TheSonOfThomp: Yeah I need to bump these back to `@latest`

utils/getComponentStory.ts
TheSonOfThomp: We should be consistent with the name, regardless of which one we use

components/pages/example/Knob.tsx
TheSonOfThomp: I'd likely rename it to something like `SBArgTypes`, since it comes from the `argTypes` property on Storybook.Meta

utils/getComponentStory.ts
spark33: Has the standardization already been done? Or will this be breaking for a little while longer for files without names that fit?

If we haven't standardized yet - I feel like `stories.tsx` would be a better name. Ignore if we've already standardized to `story.tsx`.

spark33 (approved):

utils/getComponentStory.ts
TheSonOfThomp: We're currently using `story.tsx` for everything (except leafygreen-provider)

tsconfig.json
```diff
@@ -29,5 +29,8 @@
     "utils/*.tsx",
     "pages/*.tsx"
   ],
-  "exclude": ["node_modules", "deprecated"]
+  "exclude": [
+    // "node_modules",
+    "deprecated"
+  ]
```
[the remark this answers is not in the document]
TheSonOfThomp: Yeah good catch. I needed to do this to debug, but it should be there

PR #154, opened 2022-09-19
QA Fixes

spark33:


layouts/ComponentLayout.tsx
```diff
@@ -103,80 +107,79 @@ function ComponentLayout({
...
+            <IconButton
+              // @ts-expect-error
+              as="a"
+              href={componentFields.figmaUrl}
+              target="_blank"
+              rel="noopener noreferrer"
+            >
+              <FigmaIcon />
+            </IconButton>
+          </div>
```
TheSonOfThomp, line 179: Small nit, but the weights of these icons feels off. The GH one is a lot visually heavier than the Figma one

TheSonOfThomp (approved):
LGTM

layouts/ComponentLayout.tsx
TheSonOfThomp: Screen Shot 2022-09-20 at 4 36 02 PM

spark33: That's a totally fair callout - I was thinking the same, but didn't really concern myself with it at the moment because the resolution wasn't clear. I guess I can try to make it smaller, or change the opacity on it.

I'll be making adjustments with Rob as he revisits my fixes, and I'll bring this up with him then!

components/ContentfulRichText/ContentfulRichText.tsx
```diff
@@ -1,3 +1,7 @@
+/** @jsxRuntime classic */
+/** @jsx jsx */
+// @ts-ignore unused import
+import { jsx } from '@emotion/react'
```
TheSonOfThomp, line 4: Are we using this import?

TheSonOfThomp: i.e. Are do we need to use emotion css prop or can we just use `className` like we do elsewhere?

TheSonOfThomp (approved):

PR #157, opened 2022-09-21
PD-2270: Repository Dispatch + Revalidate option on SSR

spark33:


scripts/update-packages.ts
```diff
@@ -0,0 +1,64 @@
...
+        `\n✅ Committed updated package versions to git.`,
+      )
+    );
+  }
+
+} catch (error) {
+  console.error(error)
+}
+
+export { }
```
TheSonOfThomp, line 64: We shouldn't need an empty export here

utils/getStaticComponent.ts
```diff
@@ -19,5 +19,6 @@ export async function getStaticComponentProps({ params }) {
     props: {
       component: await getComponent(params.componentName), // this is in kebabCase
     },
+    revalidate: 300, // in seconds
```
TheSonOfThomp, line 22: What does this do?

scripts/update-packages.ts
```diff
@@ -0,0 +1,64 @@
...
+  Object.prototype.hasOwnProperty.call(obj, 'version');
+const isValidUpdatesArray = (arr: any): arr is Array<ComponentUpdateObject> =>
+  Array.isArray(arr) && arr.every(isComponentUpdateObject);
+
+function exists(arg?: string | Array<any>) {
+  return arg !== undefined && arg.length > 0;
+}
+
+try {
+  const updatesArray = process.argv[2] ? JSON.parse(process.argv[2]) : [];
```
TheSonOfThomp, line 21: I'd recommend using `Commander` to do arg parsing

scripts/update-packages.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  if (exists(updatesArray) && isValidUpdatesArray(updatesArray)) {
+    updatesArray.forEach(({ name, version }: ComponentUpdateObject) => {
+      let { stdout, error } = spawnSync('yarn', ['upgrade', `${name}@^${version}`])
```
TheSonOfThomp, line 25: Since we're looping here it may be worth using the asynchronous `spawn` function. That said we'd need to use promisify, and wait for them all to resolve before moving on to the next step. Not necessary, but would be a bit cleaner

scripts/update-packages.ts
```diff
@@ -0,0 +1,64 @@
...
+    console.log(
+      chalk.green.bold(
+        `\n✅ Committed updated package versions to git.`,
+      )
+    );
+  }
+
+} catch (error) {
+  console.error(error)
+}
```
TheSonOfThomp, line 62: I don't think this will actually catch anything since you handle all the spawnSync errors

TheSonOfThomp (approved):
Some small suggestions, but nothing to hold it back

PR #158, opened 2022-09-21
Pd 2357 qa fixes

spark33:


components/ContentfulRichText/ContentfulRichText.tsx
```diff
@@ -24,10 +25,14 @@ import {
...
 
+const childrenAreNotEmpty = (children) => (
+  React.Children.count(children) && (children.toString()).trim() !== ''
+)
+
 const ContentfulRichText = ({ document }) => (
   documentToReactComponents(document, {
     renderNode: {
-      [BLOCKS.PARAGRAPH]: (node, children) => <Body css={GlobalStyles}>{children}</Body>,
+      [BLOCKS.PARAGRAPH]: (node, children) => childrenAreNotEmpty(children) && <Body css={GlobalStyles}>{children}</Body>,
```
TheSonOfThomp, line 35: Does mapping this to `undefined` in some cases not lead to errors?

TheSonOfThomp: Curious where the issue was coming from

components/ContentfulRichText/EmbeddedAsset.tsx
```diff
@@ -10,6 +10,7 @@ import { GlobalStyles } from './styles';
 const ImageWrapper = styled.div`
   ${GlobalStyles}
   max-width: 100%;
+  width: -webkit-fill-available;
```
TheSonOfThomp, line 13: Will we also need `-moz-available` and `fill-available`?

TheSonOfThomp: Does `max-content` or `fill-content` work with images?

components/ContentfulRichText/EmbeddedAsset.tsx
```diff
@@ -24,7 +25,13 @@ const EmbeddedAsset = (node) => {
 
   switch (mimeGroup) {
     case 'image':
-      return <ImageWrapper css={{ width }}><Image alt={title} src={prependUrl(file.url)} layout="responsive" width={width} height={height} /></ImageWrapper>;
+      return (
+        <ImageWrapper
+        // css={{ width }}
```
TheSonOfThomp, line 30: remove comment

components/ContentfulRichText/HorizontalLayout.tsx
```diff
@@ -5,9 +5,9 @@ const FlexContainer = styled('div')`
   display: flex;
   align-items: stretch;
   gap: 32px;
-  > * > * {
-    // first element inside column
-    width: 100%;
+  * {
+    max-width: 100%;
+    width: -webkit-fill-available;
```
TheSonOfThomp, line 10: I think babel will add this if we use `width: fill-available`

components/Footer.tsx
```diff
@@ -126,7 +127,7 @@ function Footer() {
           GitHub
         </FooterLink>
         <FooterLink href="https://www.mongodb.com/careers">Careers</FooterLink>
-        <p className={trademarkStyle}>© 2021 MongoDB, Inc.</p>
+        <p className={trademarkStyle}>© {new Date().getFullYear()} MongoDB, Inc.</p>
```
TheSonOfThomp, line 130: Perpetual copyright!!

layouts/ComponentLayout.tsx
```diff
@@ -54,18 +57,60 @@ const codeDocsWrapper = css`
...
+    width: 100%;
+    max-width: 100%;
+    overflow-x: scroll;
+  }
+`;
+
+const ComponentLinks = ({ componentFields, ...rest }) => (
+  <div {...rest}>
+    <IconButton
+      // @ts-expect-error
```
TheSonOfThomp, line 91: Should remove the `expect-error` and resolve it (I think it's asking for an `aria-label`)

pages/index.tsx
```diff
@@ -26,7 +26,7 @@ const layoutProperties = css`
     '100%',
     `${pageContainerWidth.dataGraphic}px`,
   ],
-  marginLeft: [`-${spacing[4]}px`, '0px', '0px', '0px'],
+  // marginLeft: [`-${spacing[4]}px`, '0px', '0px', '0px'],
```
TheSonOfThomp, line 29: remove commented code

styles/globals.tsx
```diff
@@ -1,9 +1,12 @@
...
+import { css } from '@emotion/css';
+import { css as EmotionReactCss } from '@emotion/react';
+import { spacing } from '@leafygreen-ui/tokens';
+import { mq } from 'utils/mediaQuery';
 import { CDN } from 'utils/routes';
 
 const fontsURL = `${CDN}/fonts`;
 
-export const globalStyles = css`
+export const globalStyles = EmotionReactCss`
```
TheSonOfThomp, line 9: Why are we importing 2 instances of `css`?

TheSonOfThomp (changes requested):
Overall looks good 👍 
Just a few small issues.

components/ContentfulRichText/ContentfulRichText.tsx
spark33: The issue was that there were empty `<p>` tags rendering, leading to empty spaces that shouldn't have been appearing.

components/ContentfulRichText/EmbeddedAsset.tsx
spark33: Thanks for catching that! I'll add it in. I'll also try those properties you listed.

components/Footer.tsx
spark33: Haha yeah I thought about just hard coding this to 2022 as it's usually done, but it was a request from Rob and I assume we'll continue to update Contentful, so I assume it was a valid move

styles/globals.tsx
spark33: IIRC, there was a TS issue with someone components that use `css` from `@leafygreen/emotion` that also need to import this `globalStyles` string. It's a weird one for sure, but this was the resolution I found outside of using `&`  at the config level that enables the `css` prop to join the different types for the separate instances of `css`.

components/ContentfulRichText/ContentfulRichText.tsx
spark33: Also - I double checked that the `renderNode` function is designed to expect `undefined`! 
image

components/Footer.tsx
```diff
@@ -1,10 +1,10 @@
-import { css } from '@emotion/css';
-import { cx } from '@leafygreen-ui/emotion';
+import { cx, css } from '@emotion/css';
```
TheSonOfThomp, line 1: Should we be using our own emotion import?

components/navigation/NavigationContent.tsx
```diff
@@ -97,7 +97,7 @@ function NavigationContent({
                 onClick={() =>
                   router.push(
                     `/component/${componentKebabCaseName}/${
-                      activeTab ?? 'example'
+                      activeTab ? activeTab : 'example'
```
TheSonOfThomp, line 100: Is this not equivalent?

pages/component/[componentName]/documentation.tsx
```diff
@@ -6,16 +6,28 @@ import { getComponent } from 'utils/getContentfulResources';
...
 const ComponentDocumentation = ({ component, changelog, readme, tsDoc }) => {
   return (
-    <CodeDocs
-      componentName={component.fields.name}
-      componentKebabCaseName={kebabCase(component.fields.name)}
-      changelog={changelog}
-      readme={readme}
-      tsDoc={JSON.parse(tsDoc) as Array<CustomComponentDoc>}
-    />
+    <div
```
Comment (TheSonOfThomp) [inline pages/component/[componentName]/documentation.tsx:15]: Does this need to be in this file? Could this be in Layout?

pages/component/[componentName]/example.tsx
```diff
@@ -28,7 +29,17 @@ const getExampleFile = component =>
 const ComponentExample = ({ component }) => {
   // todo: replace with Storybook generated live examples
   const ExampleFile = getExampleFile(component);
-  return <ExampleFile />;
+  return (
+    <div
+      className={css`
+        * {
```
Comment (TheSonOfThomp) [inline pages/component/[componentName]/example.tsx:35]: We should avoid using wildcard child selectors

pages/component/[componentName]/guidelines.tsx
```diff
@@ -12,7 +13,9 @@ const ComponentGuidelines = ({ component }) => {
   return !guidelines || guidelines.content.length == 0 ? (
     <ComingSoon />
   ) : (
-    <ContentfulRichText document={guidelines} />
+    <div className={containerPadding}>
```
Comment (TheSonOfThomp) [inline pages/component/[componentName]/guidelines.tsx:16]: Why does each page need different wrapper styling?

styles/globals.tsx
TheSonOfThomp: I'd try and resolve this. It seems like this might cause more problems down the line

TheSonOfThomp (approved):
There are still a few small issues. I'll approve now, & feel free to merge once those are addressed

components/navigation/NavigationContent.tsx
spark33: Yeah I ran into an instance where `activeTab === ''` and it caused some issues.

pages/component/[componentName]/documentation.tsx
Comment (spark33) [reply pages/component/[componentName]/documentation.tsx:15]: Not all tabs are going to have the same padding, so we'd need separate definitions.

styles/globals.tsx
spark33: If I'm not mistaken, trying to resolve this would mean either touching the `@leafygreen-ui/emotion` package so it plays nicely with `@emotion/...`, or moving all components on the website to `@emotion/...` to we can retire the `@leafygreen-ui/emotion` import. I'm not even sure the latter approach would work, since things could get messed inside LG components.

pages/component/[componentName]/guidelines.tsx
Comment (spark33) [reply pages/component/[componentName]/guidelines.tsx:16]: They'll have different values for padding!

pages/component/[componentName]/example.tsx
Comment (spark33) [reply pages/component/[componentName]/example.tsx:35]: I believe this was to fix an issue where component instances would overflow the Live Example container, so I think it might be appropriate in this situation!

components/Footer.tsx
spark33: I'll try out the other import. You're right, we should be using `@emotion/...` if it doesn't cause any issues.

PR #161, opened 2022-09-26
adds google analytics tracking

spark33:


pages/_document.tsx
```diff
@@ -68,6 +68,24 @@ export default class AppDocument extends Document {
...
 
           <link rel="icon" href="/favicon.ico" />
+
+          {/* Global Site Tag (gtag.js) - Google Analytics */}
+          <script
+            async
+            src={`https://www.googletagmanager.com/gtag/js?id=${process.env.NEXT_PUBLIC_GOOGLE_ANALYTICS}`}
+          />
+          <script
+            dangerouslySetInnerHTML={{
```
TheSonOfThomp, line 78: 😬

utils/googleAnalytics.ts
```diff
@@ -0,0 +1,13 @@
...
+  window.gtag('config', process.env.NEXT_PUBLIC_GOOGLE_ANALYTICS, {
+    page_path: url,
+  })
+}
+
+// log specific events happening.
+export const event = ({ action, params }) => {
+  // @ts-expect-error
+  window.gtag('event', action, params)
+}
```
TheSonOfThomp, line 13: is `event` ever used?

.env.example
```diff
@@ -1,4 +1,5 @@
 CONTENTFUL_ACCESS_TOKEN='<Personal access token>'
 NEXT_PUBLIC_CONTENTFUL_SPACE_ID='s9zggw79q9w3'
 NEXT_PUBLIC_CONTENTFUL_DELIVERY_TOKEN='<Content Delivery API - access token>'
-NEXT_PUBLIC_CONTENTFUL_PREVIEW_TOKEN='<Content Preview API - access token>'
\ No newline at end of file
+NEXT_PUBLIC_CONTENTFUL_PREVIEW_TOKEN='<Content Preview API - access token>'
+NEXT_PUBLIC_GOOGLE_ANALYTICS='<Tracking_ID>'
```
TheSonOfThomp, line 5: Will there be an error if this is not added?

TheSonOfThomp (approved):

.env.example
spark33: There should just be a console warning I believe!

utils/googleAnalytics.ts
spark33: I think it might not be used today, but I've been discussing potentially doing more with Google Analytics with Dave, so I figured I'd keep it here for reference.

pages/_document.tsx
spark33: This should be fine here in terms of XSS protection since we're not handling any user input!

PR #164, opened 2022-09-27
Rollback to React 17

TheSonOfThomp:
Most LG components don't support React 18 yet. We probably shouldn't upgrade yet


spark33:
Closing based on our discussion. We'll want to make sure we're catching any component-level issues with having the website run on React 18 so the entire website doesn't crash when the component instance in our `LiveExample` has issues, but using React 18 ourselves should help us ensure that all of our components are React 18-friendly.

PR #170, opened 2022-10-12
LG-2008 dynamic react example

TheSonOfThomp:


components/pages/example/LiveExample.tsx
```diff
@@ -79,7 +180,9 @@ export const LiveExample = ({
       .then(module => {
         if (module) {
           const { default: meta, ...stories } = module;
-          const defaultStoryName = meta?.parameters?.default;
+
+          const defaultStoryName =
+            meta.parameters?.default ?? Object.keys(stories)[0];
```
spark33, line 185: Is there a chance the `Object.keys(stories)[0]` fails in cases where `stories` is empty?

components/pages/example/LiveExample.tsx
```diff
@@ -110,40 +213,80 @@ export const LiveExample = ({
...
       darkMode={darkMode}
       className={css`
         margin-block: 2em;
       `}
     >
-      <div className={storyWrapperStyle}>
-        {StoryFn ? <StoryFn {...knobValues} /> : <H2>No example found 🕵️</H2>}
+      <div className={liveExampleWrapperStyle}>
+        <div className={storyWrapperStyle}>
+          {StoryFn ? <StoryFn {...knobValues} /> : <H2>No example found 🕵️</H2>}
```
spark33, line 257: 🕵️

components/pages/example/LiveExample.tsx
```diff
@@ -110,40 +213,80 @@ export const LiveExample = ({
...
+        <Transition in={showCode} timeout={200}>
+          {state => (
+            <div
+              className={cx(
+                codeExampleWrapperStyle,
+                codeWrapperStateStyle[state],
+              )}
+            >
+              <Code className={codeStyle} darkMode={darkMode} language="js">
+                {storyCode ?? 'No code found'}
```
spark33, line 268: 🕵️? 😂

spark33: Or actually, should the `No code found` text just be a Typography component outside the Code component?

components/pages/example/LiveExample.tsx
TheSonOfThomp: It gets caught in the `try/catch` and renders "No Story Found". Check out `palette`

spark33 (approved):

components/pages/example/LiveExample.tsx
TheSonOfThomp: I think it makes sense to have the text inside the code component 
Screen Shot 2022-10-20 at 1 21 40 PM

PR #172, opened 2022-10-13
on-main-push action

spark33:


TheSonOfThomp (approved):

.github/workflows/on-main-push.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+  create-pr-to-staging:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: pull-request
+      uses: repo-sync/pull-request@v2
+      with:
+        destination_branch: "staging"
+        pr_title: "Automated PR: main to staging"
```
TheSonOfThomp, line 15: Wait, why do we need this? Main should always be ahead of staging, no?

TheSonOfThomp (commented):
Not sure why we need this

.github/workflows/on-main-push.yml
spark33: There were conflicts because of the merge commit itself from staging to main. Like we discussed last week, I think a rebase from staging to main would be better so this isn't necessary. I'll close this PR!

PR #175, opened 2022-11-09
Updates `update` script

TheSonOfThomp:
Changes to `update-packages` script

- Uses `Commander` to handle args & options
- Updates all packages to `@latest` if no JSON arg is provided (useful for manually updating)
- Adds `-c`/`--commit` flag to select whether or not to commit the changes (defaults `false`)
- Adds `-v`/`--verbose` flag to print all updates


spark33 (approved):
Looks lit!

PR #176, opened 2022-11-10
Contentstack Migration

spark33:


components/ContentstackRichText/componentMap.tsx
```diff
@@ -0,0 +1,214 @@
...
+        padding: 8px;
+        vertical-align: middle;
+        border-bottom: 3px solid ${palette.gray.light2};
+      `}
+    >
+      <ContentstackChildren nodeChildren={node.children} />
+    </th>
+  ),
+  [BLOCKS.TABLE_CELL]: node => (
+    <td
```
brookescarlett, line 201: such a nit but could add a TODO here (or ticket?) to upgrade to our table components once the migration work is complete

components/navigation/NavigationContent.tsx
```diff
@@ -59,60 +64,59 @@ function NavigationContent({
...
+                    }
+                    active={contentPage.title === activePage}
+                  >
+                    {contentPage.title}
+                  </SideNavItem>
+                ))}
+            </SideNavGroup>
+          ))}
+          <SideNavGroup header="Components" glyph={<Icon glyph="Apps" />}>
+            {components
```
brookescarlett, line 94: should we do this sort higher up so that we can use the same order in both the mobile and desktop version of the navigation?

utils/getContentstackResources.ts
```diff
@@ -0,0 +1,88 @@
...
+export async function getComponent(componentName: string): Promise<any> {
+  try {
+    const query = Stack.ContentType('component').Query();
+    const result = await query
+      .where('title', startCase(componentName))
+      .includeEmbeddedItems()
+      .toJSON()
+      .find();
+    return result[0][0];
+    // const cdaUrl = `https://cdn.contentstack.io/v3/content_types/component/entries?environment=main&query=\{\"title\":\"${startCase(componentName)}\"\}&includeEmbeddedEntries[]=BASE`
```
brookescarlett, line 34: rm?

components/ContentstackRichText/componentMap.tsx
spark33: Sooa designed these guideline tables while Alven's still working on the Table component, so not 100% certain they're identical? We can create a ticket once we've confirmed.

components/ContentstackRichText/HeaderContent.tsx
```diff
@@ -45,16 +44,16 @@ const StyledIcon = styled(Icon)`
...
-      </StyledAnchor>
-    </Link>
-  );
-};
+const HeaderContent = ({
+  children,
+  headerId,
+}: PropsWithChildren<{ headerId: string }>) => (
+  <Link href={`#${headerId}`} passHref>
+    <StyledAnchor id={headerId}>
```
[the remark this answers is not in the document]
spark33: Nope. We want these links to look like the headers that wrap them, so we just want NextJS Links here.

tsconfig.tsbuildinfo
```diff
@@ -1 +1 @@
...
...2,1],[707,1],[708,1],[712,1],[714,1],[715,1],[719,1],[723,1],[724,1],[725,1],[726,1],[727,1],[728,1],[729,1],[730,1],[735,1],[736,1],[738,1],[680,1],[681,1],[682,1],[675,1],[678,1],[684,1],[691,1],[701,1],[710,1],[720,1],[721,1],[732,1],[742,1],[706,1],[663,1],[672,1],[674,1],[683,1],[686,1],[688,1],[694,1],[695,1],[696,1],[704,1],[711,1],[718,1],[722,1],[734,1],[740,1],[741,1],[660,1],[661,1],[662,1],[665,1],[666,1],[667,1],[668,1],[669,1],[671,1],[705,1],[713,1],[716,1],[731,1],[733,1],[737,1],[697,1],[685,1],[687,1],[670,1],[703,1],[709,1],[717,1],[664,1],[659,1],[679,1],[739,1],[369,1],[368,1],[643,1],[446,1],[413,1],[412,1],[445,1],[404,1],[396,1],[411,1],[408,1],[409,1],[410,1],[406,1],[407,1],[560,1],[559,1],[557,1],[616,1],[556,1],[599,1],[598,1],[444,1],[443,1],[544,1],[825,1],[551,1],[549,1],[550,1],[548,1],[405,1],[499,1],[2,1],[3,1],[4,1],[5,1],[6,1],[7,1],[8,1],[9,1],[10,1],[367,1],[402,1],[398,1],[401,1],[400,1],[399,1],[534,1],[750,1],[656,1],[657,1],[752,1],[753,1],[754,1],[749,1],[535,1],[375,1],[361,1],[751,1],[536,1],[538,1],[617,1],[539,1],[289,1],[540,1],[653,1],[541,1],[542,1],[543,1],[561,1],[360,1],[562,1],[563,1],[354,1],[370,1],[371,1]]},"version":"4.9.3"}
```
TheSonOfThomp, line 1: Don't think we should be committing this file

.env.example
```diff
@@ -1,5 +1,3 @@
-CONTENTFUL_ACCESS_TOKEN='<Personal access token>'
-NEXT_PUBLIC_CONTENTFUL_SPACE_ID='s9zggw79q9w3'
-NEXT_PUBLIC_CONTENTFUL_DELIVERY_TOKEN='<Content Delivery API - access token>'
-NEXT_PUBLIC_CONTENTFUL_PREVIEW_TOKEN='<Content Preview API - access token>'
+NEXT_PUBLIC_CONTENTSTACK_API_KEY='apikey'
+NEXT_PUBLIC_CONTENTSTACK_DELIVERY_TOKEN='token'
```
TheSonOfThomp, line 2: Could we add some details to the Readme on how to get these tokens?

components/ContentstackRichText/HeaderContent.tsx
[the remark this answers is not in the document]
TheSonOfThomp: Small nit, but should we import this as `NextLink` to avoid confusion later on?

components/ContentstackRichText/ContentstackRichText.tsx
```diff
@@ -0,0 +1,22 @@
...
+  content: any;
+  options?: any;
+}) => {
+  if ('text' in content) {
+    return <ContentstackText node={content} />;
+  } else {
+    return content.type && componentMap[content.type] ? (
+      componentMap[content.type](content, options)
+    ) : (
+      <>Unknown node type: {JSON.stringify(content)}.</>
```
TheSonOfThomp, line 17: Probably won't happen often, but should we really be outputting this to the DOM? I guess it makes it obvious if there's an issue 😆

components/ContentstackRichText/ContentstackEntry.tsx
```diff
@@ -0,0 +1,95 @@
...
+        </Badge>
+      );
+
+    case 'button_block':
+      return (
+        <Button
+          variant={entry.variant}
+          href={entry.link}
+          css={css`
+            margin: 0; // remove default Safari margin
```
TheSonOfThomp, line 44: Seems like something we should be removing in the component itself

components/ContentstackRichText/ContentstackImage.tsx
```diff
@@ -0,0 +1,33 @@
+import { css } from '@emotion/react';
+import Image from 'next/image';
+import { useState } from 'react';
+
+const ContentstackImage = ({ content }) => {
+  const attrs = content.attrs;
+  const [width, setWidth] = useState<number>(attrs['width'] ?? 700);
+  const [height, setHeight] = useState<number>(attrs['height'] ?? 300);
```
TheSonOfThomp, line 8: Are `700` & `300` just arbitrary, or are they a Contentstack default?

components/ContentstackRichText/HeaderContent.tsx
```diff
@@ -45,16 +44,16 @@ const StyledIcon = styled(Icon)`
...
-  return (
-    <Link href={`#${headerId}`} passHref>
-      <StyledAnchor id={headerId}>
-        <LinkContent>{children}</LinkContent>
-        <StyledIcon glyph="Link" fill={palette.gray.light1} />
-      </StyledAnchor>
-    </Link>
-  );
-};
+const HeaderContent = ({
```
TheSonOfThomp, line 47: Might be worth adding some short docs on what this component does

components/ContentstackRichText/componentMap.tsx
```diff
@@ -0,0 +1,217 @@
...
+            td:nth-of-type(${index + 1}),
+            th:nth-of-type(${index + 1}) {
+              width: ${colWidth}px;
+            }
+          `,
+            )}
+          `}
+        >
+          <ContentstackChildren nodeChildren={node.children} />
+        </table>
```
TheSonOfThomp, line 150: Why not our `Table`?

components/ContentstackRichText/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  ANCHOR = 'a',
+  REFERENCE = 'reference',
+  HEADING_1 = 'h1',
+  HEADING_2 = 'h2',
+  HEADING_3 = 'h3',
+  HEADING_4 = 'h4',
+  HEADING_5 = 'h5',
+  HEADING_6 = 'h6',
+  OL_LIST = 'ol',
+  UL_LIST = 'ul',
```
TheSonOfThomp, line 17: Ordered list list 😛

pages/component/[componentName]/documentation.tsx
```diff
@@ -47,9 +44,7 @@ const ComponentDocumentation = ({
 
 ComponentDocumentation.getLayout = function getLayout(page: ReactElement) {
   return (
-    <ComponentLayout componentFields={page.props.fields}>
-      {page}
-    </ComponentLayout>
+    <ComponentLayout component={page.props.component}>{page}</ComponentLayout>
```
Comment (TheSonOfThomp) [inline pages/component/[componentName]/documentation.tsx:47]: We should just make sure we're not sending too much data to the componentLayout page

pages/component/[componentName]/documentation.tsx
```diff
@@ -60,20 +55,8 @@ export async function getStaticProps({ params: { componentName } }) {
...
-    props: {
-      componentName,
-      fields,
-      changelog,
-      readme,
-      tsDoc,
-    },
-  };
+  const component = await getComponent(componentName);
+  return { props: { componentName, component, changelog, readme, tsDoc } };
```
Comment (TheSonOfThomp) [inline pages/component/[componentName]/documentation.tsx:59]: Did this just get prettified?

pages/component/[componentName]/example.tsx
```diff
@@ -18,27 +18,21 @@ const ComponentExample = ({
...
 export const getStaticPaths = getStaticComponentPaths;
 
 export const getStaticProps = async ({ params }) => {
   const { componentName } = params;
-
-  // Here we pull out the designGuidelines to ensure we're not passing that to this page unnecessarily
-  const { designGuidelines, ...fields } = await getComponentFields(
-    componentName,
-  );
+  const component = await getComponent(componentName);
```
Comment (TheSonOfThomp) [inline pages/component/[componentName]/example.tsx:29]: Similar comment to above. We should make sure we're not sending _all_ the data to the `example` & `docs` pages unnecessarily

utils/getContentstackResources.ts
```diff
@@ -0,0 +1,76 @@
...
+  api_key: process.env.NEXT_PUBLIC_CONTENTSTACK_API_KEY as string,
+  delivery_token: process.env.NEXT_PUBLIC_CONTENTSTACK_DELIVERY_TOKEN as string,
+  environment: 'main',
+});
+
+export async function getComponents(): Promise<any> {
+  try {
+    const results = await Stack.ContentType('component')
+      .Query()
+      .limit(100)
```
TheSonOfThomp, line 14: What happens when we reach 100 components? 😛

utils/getContentstackResources.ts
```diff
@@ -0,0 +1,76 @@
...
+    const result = await query
+      .where('title', startCase(componentName))
+      .includeEmbeddedItems()
+      .toJSON()
+      .find();
+    return result[0][0];
+  } catch (error) {
+    console.error(error);
+  }
+}
```
TheSonOfThomp, line 37: Should maybe consider being more explicit in the error message

utils/getContentstackResources.ts
```diff
@@ -0,0 +1,76 @@
...
+    const query = Stack.ContentType('content_page').Query();
+    const result = await query
+      .where('title', startCase(contentPageName))
+      .includeEmbeddedItems()
+      .toJSON()
+      .find();
+    return result[0][0];
+  } catch (error) {
+    console.error(error);
+  }
```
TheSonOfThomp, line 65: Same here

components/ContentstackRichText/ContentstackRichText.tsx
spark33: Yeah I was thinking since the users are mainly MongoDB designers, being able to diagnose issues faster would be easier. We could also explore having a component to render bugs like this in a more user-friendly way in the future.

components/ContentstackRichText/ContentstackEntry.tsx
spark33: Feel free to create a ticket if you think so! I think I added this because Safari's browser default styles adds a margin.

components/ContentstackRichText/ContentstackImage.tsx
spark33: 700 is because the `max-width` of the container is 700px, and that's what most of our guidelines assets are optimized for. 300 is arbitrary - most of our images seemed to render decently there.

These values should only be defined when everything that can go wrong goes wrong though. Didn't think it was a big deal.

components/ContentstackRichText/componentMap.tsx
spark33: The structure of the rich text JSON doesn't play well with our current API. We could migrate after the new Table component is build though

components/ContentstackRichText/types.ts
spark33: LOL that's how contentful had it and I never changed it - changing it now.

pages/component/[componentName]/example.tsx
Comment (spark33) [reply pages/component/[componentName]/example.tsx:29]: Yeah I agree. I think at this point though, we consider it a future performance improvement and not let it block this merge. Partially because I just want to get it in, but also because there are more ways to optimize the performance here now that we're using Contentstack. For example, I believe Contentstack allows us to make requests for entries that just exclude certain fields. It would be nice to identify performance improvements and do a full sweep.

tsconfig.tsbuildinfo
spark33: I don't know why it was committed when it's already marked in `.gitignore`. Removed!

pages/component/[componentName]/documentation.tsx
Comment (spark33) [reply pages/component/[componentName]/documentation.tsx:59]: No, I removed it when I was first starting out with dev on this migration just to reduce the number of variables that could introduce bugs. More on this on a later comment

utils/getContentstackResources.ts
spark33: Removed! It was just something I tried when the request wasn't working.

tsconfig.tsbuildinfo
TheSonOfThomp: Looks like it was added again. Might need to follow this: https://stackoverflow.com/a/7532131/2200383

components/ContentstackRichText/ContentstackEntry.tsx
```diff
@@ -0,0 +1,95 @@
...
+          `}
+        >
+          {entry.content}
+        </Button>
+      );
+
+    case 'callout_block':
+      return (
+        <Callout variant={Variant[entry.variant]} className="nested-entry">
```
brookescarlett, line 53: why className rather than css here?

components/ContentstackRichText/ContentstackImage.tsx
TheSonOfThomp: We should avoid using magic numbers here though. If it's the max-width of the container, we should use the same variable

components/ContentstackRichText/componentMap.tsx
```diff
@@ -0,0 +1,217 @@
...
+import HeaderContent from './HeaderContent';
+import { BLOCKS } from './types';
+import { palette } from '@leafygreen-ui/palette';
+
+const componentMap = {
+  [BLOCKS.DOCUMENT]: (node, options) => {
+    return (
+      <div
+        css={
+          (!options || !options.isNested) &&
```
brookescarlett, line 23: isNested feels very heavy to me. We're passing it through options in a lot of places only to change how margin is set. Does it make sense to pass this prop everywhere or just set the css where relevant? Not blocking feedback, just thinking outloud

components/ContentstackRichText/ContentstackEntry.tsx
TheSonOfThomp: 👍 https://jira.mongodb.org/browse/LG-2681

brookescarlett: Leave a comment on that ticket to clean this up after that's fixed :)

pages/component/[componentName]/documentation.tsx
Comment (TheSonOfThomp) [reply pages/component/[componentName]/documentation.tsx:59]: Should run `fix` then 😛

pages/component/[componentName]/example.tsx
Comment (TheSonOfThomp) [reply pages/component/[componentName]/example.tsx:29]: That's fair. We should make sure we don't forget about this though

Comment (brookescarlett) [reply pages/component/[componentName]/example.tsx:29]: Do we have a perf ticket? if not, let's make one? And maybe add to the .design bug bash doc? I want to keep track of things we find blocking v. not blocking before releasing to main.

utils/getContentstackResources.ts
TheSonOfThomp: Elsewhere we have `console.error('No Component pages found', error);`. We should add a short message here saying "Component ${componentName } not found"

components/ContentstackRichText/ContentstackImage.tsx
brookescarlett: ➕

components/ContentstackRichText/ContentstackEntry.tsx
TheSonOfThomp: I think we typically use `className`. I'd ask why `css` elsewhere 😛

components/ContentstackRichText/componentMap.tsx
spark33: The tricky thing is that all nested entries in Contentstack come to us as `type === BLOCKS.Document`, so it's tricky to define the margin at the document level. We could just override this margin at the page level maybe? But I thought this would be a pretty flexible pattern that we could use to support other options in the future?

brookescarlett: Kk you may be right, re future proofing

brookescarlett (approved):

brookescarlett:
Make sure @TheSonOfThomp is all set before merging :)

components/ContentstackRichText/ContentstackEntry.tsx
spark33: I think `css` is the preferred prop for css strings in the latest emotion? Let me know if I'm wrong here.

We use className in leafygreen-ui because we're on an old emotion version, but I think the new emotion pushes us to use either styled components or the css prop.

pages/component/[componentName]/documentation.tsx
Comment (spark33) [reply pages/component/[componentName]/documentation.tsx:59]: Oh you were just talking about why they're on one line! I thought you were talking about why designGuidelines was still in these props. My bad. I did run `fix`, but this is what comes out! Don't ask me why LOL

PR #185, opened 2023-01-06
Update packages

TheSonOfThomp:
Updates LG packages
Also updates `isPathInLeafygreen` regex to include _all_ `node_modules/@leafygreen-ui` packages in babel transpilation


package.json
```diff
@@ -27,51 +27,51 @@
     "@emotion/server": "^11.0.0",
     "@emotion/styled": "^11.9.3",
     "@leafygreen-ui/a11y": "^1.3.4",
-    "@leafygreen-ui/badge": "^8.0.0",
-    "@leafygreen-ui/banner": "^7.0.0",
+    "@leafygreen-ui/badge": "^8.0.1",
```
brookescarlett, line 30: are we just running `yarn upgrade` or is this manual?

TheSonOfThomp: Effectively, yes. We have a script that runs `yarn upgrade ...<all LG packages>`

https://github.com/mongodb/design/blob/staging/scripts/update-packages.ts

brookescarlett: Ah got it

brookescarlett (approved):

PR #193, opened 2023-01-13
LG-2738 consolidate inherited docs

TheSonOfThomp:
Screen Shot 2023-01-13 at 12 47 13 PM


TheSonOfThomp:
> A couple of notes: 
> 
> 
> 
> - Previously we were more specific about the type of HTMLAttributes that a component was inheriting (div, when base element was div) and now that's gone away.
> 

This depends on what component you're looking at. Components that explicitly extend `div` will show that


I'll fix the semicolon 👍

brookescarlett (approved):

PR #194, opened 2023-01-13
LG-2738 guidecue docs

TheSonOfThomp:
- Adds Guide Cue to node_modules
- Updates value mapping for Enum props

### DIFF
--- a/components/pages/example/LiveExample.tsx
+++ b/components/pages/example/LiveExample.tsx
@@ -78,6 +78,8 @@ export const LiveExample = ({
             tsDoc,
           });
 
+          console.log(state);
+
           setState(state);
         } else {
           setState(initialLiveExampleState);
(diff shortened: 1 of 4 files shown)


brookescarlett:
The guidecue live example does not work and a lot of the props in the Tooltip prop table are missing their descriptions

TheSonOfThomp:
> The guidecue live example does not work and a lot of the props in the Tooltip prop table are missing their descriptions

Guidecue example: 
Screen Shot 2023-01-18 at 1 24 09 PM

You'll probably need to `yarn --force`

This PR shouldn't affect `Tooltip`

TheSonOfThomp:
> TooltipProps within Guidecue code docs @TheSonOfThomp

Ah I see. That'll have to be fixed in the TSDoc on the GuideCue `TooltipContent` component itself: https://github.com/mongodb/leafygreen-ui/blob/main/packages/guide-cue/src/TooltipContent.tsx#L30-L39

brookescarlett:
Pulled the latest and the "next" button still isn't working

TheSonOfThomp:
> Pulled the latest and the "next" button still isn't working

I think that's just how the story is written https://thesonofthomp.github.io/leafygreen-ui/?path=/story/components-guidecue--default

brookescarlett:
yep you're right.. it's weird/looks like a bug but i won't block on it

brookescarlett (approved):

TheSonOfThomp:
> yep you're right.. it's weird/looks like a bug but i won't block on it

👍 I agree it's weird. I'll file a ticket to fix the story itself

PR #202, opened 2023-01-23
LG-2684 Performance Enhancements

TheSonOfThomp:


components/NextLinker.tsx
```diff
@@ -0,0 +1,7 @@
+import NextLink from 'next/link';
+
+export const NextLinker = ({ href, children, ...props }) => (
```
brookescarlett, line 3: nit but why NextLinker would prefer NextLinkWrapper?

pages/component/[componentName]/example.tsx
```diff
@@ -25,9 +26,18 @@ ComponentExample.getLayout = function getLayout(page: ReactElement) {
...
+  params,
+}): Promise<{ props: ExamplePageProps }> => {
   const { componentName } = params;
-  const component = await getComponent(componentName);
+  const component: ComponentPageMeta | null =
+    (await getComponent(componentName, {
+      includeContent: false,
+    })) ?? null;
+
+  // Find the TSDoc object only for the primary component
```
Comment (brookescarlett) [inline pages/component/[componentName]/example.tsx:38]: rm?

brookescarlett:
should the subheader title be "Foundation" or "Foudnations"
Approving because no blocking feedback but left a few comments

components/NextLinker.tsx
TheSonOfThomp: That's just how we've documented it in other fixes

TheSonOfThomp:
> should the subheader title be "Foundation" or "Foundations" 

Not sure. This is set in the CMS though

PR #203, opened 2023-01-24
Sort imports

TheSonOfThomp:
Adds `eslint-plugin-simple-import-sort`.

Any code changes are from running `yarn fix`


brookescarlett (approved):

brookescarlett:
Can we share eslint/prettier/etc. config between the two repos?

TheSonOfThomp:
> `Can we share eslint/prettier/etc. config between the two repos?

I'd love that—not sure what the best way to do that is though

PR #205, opened 2023-01-25
Updates Contentstack types

TheSonOfThomp:
Adds types for Contentstack entries & blocks
Updates `ExampleCard` style

Screen Shot 2023-01-26 at 11 54 25 AM

### DIFF
--- a/components/ContentstackRichText/ExampleCardBlock/ExampleCardBlock.tsx
+++ b/components/ContentstackRichText/ExampleCardBlock/ExampleCardBlock.tsx
@@ -30,10 +30,15 @@ const Subtext = styled(Body)`
 
 const ExampleCardBlock = ({ entry }: { entry: ExampleCardBlockProps }) => {
   const IconComponent = Icons[entry.variant];
-
   return (
     <div>
       <ImageContainer color={BorderColors[entry.variant]}>
+        {/**
+         * TODO: fix this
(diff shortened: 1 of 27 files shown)


components/ContentstackRichText/ExampleCardBlock/ExampleCardBlock.tsx
```diff
@@ -24,24 +28,29 @@ const Subtext = styled(Body)`
...
+const ExampleCardBlock = ({ entry }: { entry: ExampleCardBlockProps }) => {
   const IconComponent = Icons[entry.variant];
   return (
     <div>
       <ImageContainer color={BorderColors[entry.variant]}>
+        {/**
+         * TODO: fix this
+         * Contentstack doesn't send image sizes,
+         * so we can't appropriately size a Next/Image component
+         * */}
```
spark33, line 40: We could use the ContentstackImage component here!

spark33 (approved):

components/ContentstackRichText/ExampleCardBlock/ExampleCardBlock.tsx
TheSonOfThomp: I'm getting an error, since the object `entry.image` returned by CS doesn't have an `attrs` property. Not sure why the API is so inconsistent

PR #217, opened 2023-02-08
Small .design tweaks

brookescarlett:
- Foundations pages were not appearing as active in side nav -- fixes that
- Increases line height of description items in PropTable

### DIFF
--- a/components/navigation/NavigationContent.tsx
+++ b/components/navigation/NavigationContent.tsx
@@ -73,18 +73,22 @@ function NavigationContent({
               glyph={<Icon glyph={contentPageGroup.iconname} />}
             >
               {contentPageGroup.content_pages &&
-                contentPageGroup.content_pages.map(contentPage => (
-                  <SideNavItem
-                    key={contentPage.title}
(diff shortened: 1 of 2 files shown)


components/navigation/NavigationContent.tsx
```diff
@@ -73,18 +73,22 @@ function NavigationContent({
...
+                contentPageGroup.content_pages.map(contentPage => {
+                  const contentPageKebabCaseName = kebabCase(contentPage.title);
+
+                  return (
+                    <SideNavItem
+                      key={contentPage.title}
+                      as={NextLinkWrapper}
+                      href={`/${kebabCase(contentPageGroup.title)}/${kebabCase(
+                        contentPage.title,
+                      )}`}
```
spark33, line 85: The `kebabCase(
                        contentPage.title,
                      )` inside the `href` prop can just be `contentPageKebabCaseName` now

brookescarlett: Good catch

PR #219, opened 2023-02-23
LG-2906: Pagination package added to package.json

spark33:


package.json
```diff
@@ -52,6 +52,7 @@
     "@leafygreen-ui/marketing-modal": "^4.0.3",
     "@leafygreen-ui/menu": "^19.0.4",
     "@leafygreen-ui/modal": "^14.0.1",
+    "@leafygreen-ui/pagination": "^1.0.1",
```
TheSonOfThomp, line 55: 👍 Want to run `update-packages` to get any other bumps as well?

TheSonOfThomp (approved):
👍

package.json
spark33: I ran `yarn` locally, so it should've gotten the other bumps as well!

PR #239, opened 2023-04-06
Table Adapter in Prop Tables

TheSonOfThomp:
Still TODO: Update table packages


spark33:
Note: The `Table props` table in Code Docs and Live Example knobs are being pulled from the V10 Table since both components have the same displayName. We may want to wait for Table v11.0.2 with an appropriate fix before this is merged.

spark33:
Not needed as a full migration to v11 is in PR

PR #246, opened 2023-04-11
Live example decorators

TheSonOfThomp:
- Adds any story decorators to the live examples
- Imports stories from `story.js`


spark33 (changes requested):
There's also a bug on the Typography live example.

spark33 (changes requested):
Found three bugs - 

1) After you go from one component's live example page to another, clicking the 'Show Code' button displays the code from the first component's live example.

2) This looks pretty awkward - feels like the error message should be below the heading.
image

3) It looks like when the live example is loading, nothing is shown.

TheSonOfThomp:
Fixes 1 & 2 👍 

> 3. It looks like when the live example is loading, nothing is shown.
That's as designed, to reduce flicker of the loading message for fast loads. The Loading screen only shows after `loadingTimeout = 1000;`

spark33 (approved):

PR #247, opened 2023-04-13
LG-2999, LG-3010: Guideline styles

spark33:


package.json
```diff
@@ -83,6 +84,7 @@
     "@storybook/react": "^6.5.10",
     "@storybook/source-loader": "^6.5.12",
     "@svgr/webpack": "^6.3.1",
+    "@tanstack/react-table": "^8.7.3",
```
brookescarlett, line 87: why do we need this

brookescarlett: shouldn't it be a dep of our table component directly?

brookescarlett (approved):

PR #258, opened 2023-04-20
LG-2036, LG-2917: ErrorBoundary + Style changes

spark33:


components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { PropsWithChildren } from 'react';
+
+class ErrorBoundary extends React.Component<
```
bruugey, line 3: why is this a class component?

spark33: It needs a specific static function. Docs [here](https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary)

bruugey (approved):

PR #259, opened 2023-04-21
Code example generator [LG-2954, LG-2950, LG-2951]

TheSonOfThomp:
Traverses the rendered example component in the ReactTree to find the relevant component, then generates the JSX string from that component.

To Test:
Link this branch: https://github.com/mongodb/leafygreen-ui/pull/1700
Rebuild packages (`yarn build`) and docs (`yarn docs:tsdoc`)


components/pages/example/LiveExample.tsx
```diff
@@ -27,7 +28,6 @@ import {
 } from './LiveExampleStateComponents';
 import {} from './types';
```
bruugey, line 29: rm?

components/pages/example/useLiveExampleState/useLiveExampleState.ts
```diff
@@ -1,5 +1,6 @@
 import { useReducer } from 'react';
 import { kebabCase, merge } from 'lodash';
+import pascalcase from 'pascalcase';
```
bruugey, line 3: is this in our packagejson? was playing around with .design recently and got an error that this packge could not be found

components/pages/example/utils/getStoryCode.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  /** Returns the component root we want to generate source code for */
+  function getComponentRoot(
+    renderedStory: FunctionComponentElement<any>,
+    componentName: string,
+  ) {
+    let isRootSet = false;
+    let componentRoot: ReactElement<any> = renderedStory;
+
+    // @ts-expect-error - prepass callback args are incorrectly typed. Need to explicitly re-type them
```
bruugey, line 47: is there an open issue on their side that we can reference here?

components/pages/example/utils/getStoryCode.ts
```diff
@@ -0,0 +1,72 @@
...
+    prepass(renderedStory, (element: ReactElement<any>) => {
+      if (!isRootSet && isFunctionComponentElement(element)) {
+        if (packageNameDoesNotMatchComponent.includes(componentName)) {
```
bruugey, line 50: should this condition be moved up a level?

components/pages/example/useLiveExampleState/useLiveExampleState.ts
TheSonOfThomp: I think it's a dependency of something else, but it should be explicit

components/pages/example/utils/getStoryCode.ts
TheSonOfThomp: No, since we still need to check `isFunctionComponentElement `

components/pages/example/utils/getStoryCode.ts
TheSonOfThomp: There is now https://github.com/FormidableLabs/react-ssr-prepass/issues/86

TheSonOfThomp: Though the last commit to that repo was over a year ago, so no guarantees it gets in. I _could_ try and find another library that does the same, but in my research, the other options were even more out-of-date. (another option would be to port over the code and maintain it ourselves, but that doesn't feel right to me)

bruugey (commented):
Will approve once Table bug is addressed!

components/pages/example/utils/getStoryCode.ts
```diff
@@ -43,13 +43,16 @@ export function getStoryCode(context: LiveExampleContext): string | undefined {
         filterProps: (value, name) => {
           const tsProp = TSPropsArray.find(p => p.name === name);
           const tsDefault = tsProp ? getDefaultValueValue(tsProp) : null;
+
+          const excludeProp =
+            ignoreAllPropsForComponents.includes(componentName) ||
+            ignoreProps.includes(name) ||
+            value === tsDefault ||
+            name === 'key';
```
bruugey, line 51: should we add key to ignoreProps?

TheSonOfThomp: 👍 Yeah

bruugey (approved):

PR #260, opened 2023-04-24
Sort Knobs table same as code docs [LG-3128]

TheSonOfThomp:
Shares the sort function between code docs & live example knobs.
Sorts required props to the top & then the rest alphabetically


components/pages/example/KnobsTable/KnobRow.tsx
```diff
@@ -31,6 +33,13 @@ const knobControlStyle = css`
   justify-content: end;
 `;
 
+const requiredFlagStyle = css`
+  display: inline;
+  padding-left: 1ch;
```
bruugey, line 38: TIL

bruugey (approved):

PR #300, opened 2023-09-19
LG-3581: Desktop login button (FE only)

spark33:


pages/.DS_Store
TheSonOfThomp, line 1: rm

localhost.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
TheSonOfThomp, line 1: 😬

pages/.DS_Store
spark33: I'm actually the one removing this with my PR hahah

localhost.key
spark33: A bit of a scare, but this is just my own key to set up SSL on localhost. Shouldn't be a security risk. Removing!

components/navigation/NavigationContent.tsx
```diff
@@ -58,7 +74,26 @@ function NavigationContent({
...
+                      )}
+                    </div>
+                    {component.private && (
+                      <Description
+                        style={{
+                          textTransform: 'none',
+                          color: palette.gray.base,
+                        }}
+                      >
+                        Log in to view this component
```
shaneeza, line 93: this looks a little tight on a small device 
Screenshot 2023-10-11 at 12 49 08 PM

components/SignIn/DesktopSignIn.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+import { UserInfo } from '../UserInfo';
+
+const Container = styled('div')`
+  width: 100%;
+  align-items: center;
+  justify-content: flex-end;
+  padding-top: ${spacing[4]}px;
+  z-index: 1;
+  // todo: remove
```
shaneeza, line 23: Can you give a little more context on this todo?

components/pages/example/LiveExample.tsx
```diff
@@ -10,7 +10,7 @@ import LeafyGreenProvider, {
 
 import { KnobsTable } from './KnobsTable/KnobsTable';
 import { isStateReady } from './useLiveExampleState/utils';
-import { CodeExample } from './CodeExample';
+// import { CodeExample } from './CodeExample';
```
shaneeza, line 13: Why all the commenting out?

layouts/ComponentLayout.tsx
```diff
@@ -148,62 +159,73 @@ function ComponentLayout({
...
-                  className={cx([flexContainer, desktopLinksContainer])}
-                />
-              )
-            }
-            as={NextLinkWrapper}
-          >
-            <Tab
-              name="Live Example"
-              href={`/component/${kebabCase(component.title)}/example`}
+        {component && session ? (
```
shaneeza, line 180: Is this saying don't show any of the tabs(example, guidelines/docs) unless there is a session? I'm not logged in and I can't see any of the tabs 
Screenshot 2023-10-11 at 1 19 06 PM

pages/_app.tsx
```diff
@@ -3,36 +3,40 @@ import { Global } from '@emotion/react';
...
 function MyApp({
   Component,
-  pageProps,
+  pageProps: { session, ...pageProps },
   components,
   contentPageGroups,
-}: AppPropsWithLayout) {
+}: AppPropsWithLayout<{ session: Session }> & {
+  components: Array<any>;
+  contentPageGroups: Array<any>;
```
shaneeza, line 38: Can you explain the reason for splitting these from the type above?

components/SignIn/DesktopSignIn.tsx
spark33: I put that gap there before sessions were implemented to separate the login and user buttons. Now that there's just one button, it can be removed! Removed :)

components/pages/example/LiveExample.tsx
spark33: This is in line with the CodeExamples being temporarily disabled. Other CodeExample-related code lines were commented, so I commented these changes out too

pages/_app.tsx
spark33: Yeah! `AppPropsWithLayout` is an export from `next`, and these props are our own

components/navigation/NavigationContent.tsx
spark33: We're exploring a single stack of navigation items to resolve this issue. @brookeyalof any qualms with me trying that approach out?

spark33: Or is this a small enough issue that we don't want to change this navigation so close to DDP

spark33:
> Since the [Side Nav](https://github.com/mongodb/design/blob/staging/layouts/BaseLayout.tsx#L72-L86) is wrapped in the leafygreen provider that has the portal and scroll container set, the tooltip won't scroll with the side nav when the side nav scrolls. https://jmp.sh/0ogMepKq. I think we would need to pass the ref of the side nav to the portalContainer prop on the tooltip but I don't even think side nav forwards a ref, or maybe, try `usePortal=false`

`usePortal={false}` doesn't render a Tooltip at all, and I was trying to see if that was `overflow: hidden` hiding it, and removing that CSS prop didn't do the trick. I'll create a ticket so `SideNav` takes in a ref.

components/SignIn/DesktopSignIn.tsx
```diff
@@ -0,0 +1,92 @@
...
+      {session ? (
+        <Menu
+          trigger={
+            <UserMenuTrigger rightGlyph={<Icon glyph="CaretDown" />}>
+              {session?.user?.name}
+            </UserMenuTrigger>
+          }
+        >
+          <FocusableMenuItem>
+            <UserInfo
```
shaneeza, line 67:
```
next-dev.js:20 Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
```
I'm seeing this error in the console and i think its because `FocusableMenuItem` is trying to [add](https://github.com/mongodb/leafygreen-ui/blob/main/packages/menu/src/MenuItem/FocusableMenuItem.tsx#L12-L14) a `ref` to `UserInfo` but `UserInfo` does not forward a ref

components/navigation/NavigationContent.tsx
```diff
@@ -58,7 +74,26 @@ function NavigationContent({
...
+                  <div>
+                    <div style={{ display: 'flex' }}>
+                      {component.title}
+                      {component.private && (
+                        <LockIconContainer>
+                          <Icon glyph="Lock" />
+                        </LockIconContainer>
+                      )}
+                    </div>
+                    {component.private && (
```
shaneeza, line 86: Screenshot 2023-10-16 at 4 36 13 PM
 im logged in but the log in text is still visible

components/SignIn/MobileSignIn.tsx
```diff
@@ -0,0 +1,52 @@
...
+`;
+
+export const MobileSignIn = () => {
+  const { data: session } = useSession();
+  return (
+    <Container>
+      {session ? (
+        <>
+          <UserInfo
+            name={session?.user?.name ?? 'User'}
```
shaneeza, line 36: Are you accounting for really long names and email addresses?

components/SignIn/DesktopSignIn.tsx
shaneeza: I'm noticing that you also can't keyboard focus into the menu and this might be why.

shaneeza:
> > Since the [Side Nav](https://github.com/mongodb/design/blob/staging/layouts/BaseLayout.tsx#L72-L86) is wrapped in the leafygreen provider that has the portal and scroll container set, the tooltip won't scroll with the side nav when the side nav scrolls. https://jmp.sh/0ogMepKq. I think we would need to pass the ref of the side nav to the portalContainer prop on the tooltip but I don't even think side nav forwards a ref, or maybe, try `usePortal=false`
> 
> `usePortal={false}` doesn't render a Tooltip at all, and I was trying to see if that was `overflow: hidden` hiding it, and removing that CSS prop didn't do the trick. I've created a [ticket](https://jira.mongodb.org/browse/LG-3687) so `SideNav` takes in a ref.

I think it is hidden because of overflow. I tried it out and the tooltip is there in the inspector. that sucks

components/SignIn/MobileSignIn.tsx
spark33: Added `word-break: break-word`, so now it should render well!

TheSonOfThomp:
I'm getting an OAuth error when trying to sign in 

`http://localhost:3000/api/auth/signin/?error=OAuthSignin`
Screenshot 2023-10-18 at 1 43 41 PM

components/Avatar.tsx
```diff
@@ -0,0 +1,17 @@
+import styled from '@emotion/styled';
+
+import { palette } from '@leafygreen-ui/palette';
+
+const Avatar = styled('div')`
+  border-radius: 50%;
+  min-width: 36px;
+  min-height: 36px;
+  background-color: ${palette.gray.base};
+  color: white;
```
shaneeza, line 10: you can use palette 🌈

layouts/ComponentLayout.tsx
```diff
@@ -99,9 +101,11 @@ const ComponentLinks = ({
       target="_blank"
       rel="noopener noreferrer"
       style={{ marginRight: '8px' }}
-      href={`https://github.com/mongodb/leafygreen-ui/tree/main/packages/${kebabCase(
-        component.title,
-      )}`}
+      href={`https://github.com/${
```
shaneeza, line 104: nit: create variables for these

components/Avatar.tsx
```diff
@@ -0,0 +1,17 @@
+import styled from '@emotion/styled';
+
+import { palette } from '@leafygreen-ui/palette';
+
+const Avatar = styled('div')`
```
TheSonOfThomp, line 5: 🔥

package.json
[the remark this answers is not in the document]
spark33: It isn't necessary here since `mongo-nav` isn't set up with `tsdoc.json` and `stories.js`. So adding the package here won't really do anything

components/SignIn/DesktopSignIn.tsx
```diff
@@ -0,0 +1,92 @@
...
+  align-items: center;
+  justify-content: flex-end;
+  padding-top: ${spacing[4]}px;
+  z-index: 1;
+  ${mq({
+    display: ['none', 'none', 'flex'],
+  })}
+`;
+
+const Menu = styled(LGMenu)`
```
TheSonOfThomp, line 28: nit: I'd maybe prefer `StyledMenu = styled(Menu)`, but it doesn't really matter

components/SignIn/DesktopSignIn.tsx
```diff
@@ -0,0 +1,92 @@
...
+            </UserMenuTrigger>
+          }
+        >
+          <FocusableMenuItem>
+            <UserInfo
+              name={session?.user?.name ?? 'User'}
+              email={session?.user?.email ?? ''}
+            />
+          </FocusableMenuItem>
+          <LogOutMenuItem
```
TheSonOfThomp, line 72: +1 to what Shaneeza mentioned. Can't keyboard select the LogOut button

components/Unauthorized.tsx
```diff
@@ -0,0 +1,34 @@
...
+const Wrapper = styled('div')`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+`;
+
+export default function Unauthorized() {
+  return (
+    <Wrapper>
+      <BasicEmptyState
```
TheSonOfThomp, line 18: this looks a little funky on mobile, but that's out-of-scope I believe

TheSonOfThomp: Screenshot 2023-10-18 at 2 42 22 PM

components/pages/example/LiveExample.tsx
TheSonOfThomp: Examples were severely broken, so until we're able to fix them, we need to disable them.
Surprised this unused import wasn't caught by ESLint

layouts/BaseLayout.tsx
```diff
@@ -84,9 +85,11 @@ function BaseLayout({ children }: { children: React.ReactNode }) {
         <LayoutContext.Provider value={bodyContainerRef}>
           <div className={containerStyle}>
             <Navigation />
-
             <div className={layout} ref={setBodyContainerRef}>
-              <div className={childrenWrapper}>{children}</div>
+              <div className={childrenWrapper}>
+                <SignIn />
+                {children}
+              </div>
```
TheSonOfThomp, line 92: Should the Login button/UserMenu trigger be in line with the page header? Everything feels like it's shifted down a lot

layouts/ComponentLayout.tsx
```diff
@@ -148,62 +159,73 @@ function ComponentLayout({
...
-          >
-            <Tab
-              name="Live Example"
-              href={`/component/${kebabCase(component.title)}/example`}
+        {component && (!component.private || (component.private && session)) ? (
+          <div className={flexContainer}>
+            <Tabs
+              selected={selected}
+              setSelected={setSelected}
+              aria-label={`Information on LeafyGreen UI ${pageTitle} component`}
```
TheSonOfThomp, line 185: should be `componentName`. Otherwise the label is set to `"Information on LeafyGreen UI Mongo Nav – LeafyGreen Design System | MongoDB component""`

TheSonOfThomp: Also, unsure if `Information on LeafyGreen UI ${componentName} component` is all that useful of a label here either

package.json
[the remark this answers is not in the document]
TheSonOfThomp: 👍 so that'll be added later then I assume

pages/_app.tsx
TheSonOfThomp: Can we be more precise than `Array<any>`

pages/api/auth/api-auth.ts
```diff
@@ -0,0 +1,24 @@
...
+import { NextApiRequest, NextApiResponse } from 'next';
+import { unstable_getServerSession } from 'next-auth/next';
+
+import { authOptions } from './[...nextauth]';
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse,
+) {
+  const session = await unstable_getServerSession(req, res, authOptions);
```
TheSonOfThomp, line 11: 😬 Is there a `stable` version?

pages/component/[componentName]/documentation.tsx
```diff
@@ -47,7 +47,12 @@ const ComponentDocumentation = ({
 
 ComponentDocumentation.getLayout = function getLayout(page: ReactElement) {
   return (
-    <ComponentLayout component={page.props.component}>{page}</ComponentLayout>
+    <ComponentLayout
+      componentName={page.props.componentName}
+      component={page.props.component}
```
Comment (TheSonOfThomp) [inline pages/component/[componentName]/documentation.tsx:52]: we should be able to pull `name` off of `component`, no?

pages/component/private/[componentName]/documentation.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { css, cx } from '@emotion/css';
+
+interface DocsPageProps {
+  componentName: string;
+  changelog: string;
+  readme: string;
+  tsDoc: Array<CustomComponentDoc>;
+}
+
+const ComponentDocumentation = ({
```
Comment (TheSonOfThomp) [inline pages/component/private/[componentName]/documentation.tsx:25]: Can we de-dupe as much as possible between the `private/` pages and regular pages?

utils/ContentStack/getContentstackResources.ts
```diff
@@ -38,7 +38,14 @@ interface QueryOptions {
...
 }
 
-const componentProperties = ['uid', 'title', 'description', 'url', 'figmaurl'];
+const componentProperties = [
+  'uid',
+  'title',
+  'description',
+  'url',
+  'figmaurl',
+  'private',
```
TheSonOfThomp, line 47: nit: `isPrivate`?

TheSonOfThomp (changes requested):
Left some comments.
We should also add the required keys to `.env.example`

layouts/BaseLayout.tsx
spark33: The login button design was done with the new Side Nav + IA in mind, so that's when it'll start looking more aligned

pages/api/auth/api-auth.ts
spark33: Using the stable version requires bumping to R18, and the unstable version was [specifically documented as being fully tested with no bugs reported](https://github.com/nextauthjs/next-auth/pull/6537/files). Essentially, when they went from `unstable` to `stable`, they did a simple rename. The `stable` one is actually new enough that all guides for next-auth online were written with `unstable`.

pages/component/[componentName]/documentation.tsx
Comment (spark33) [reply pages/component/[componentName]/documentation.tsx:52]: Not with private components! `component` won't be passed if the component is private and the user is not signed in

components/Unauthorized.tsx
spark33: Yeah this can't really be fixed until EmptyState components themselves are mobile friendly

utils/ContentStack/getContentstackResources.ts
spark33: This is what's shown on Contentstack and I thought from a non-engineering perspective, `Private` might be nicer?

pages/component/private/[componentName]/documentation.tsx
Comment (spark33) [reply pages/component/private/[componentName]/documentation.tsx:25]: I can de-dupe the types! It feels like the rendering markup is already pretty well de-duped using the `CodeDocs`, `LiveExample`, etc components.

layouts/ComponentLayout.tsx
spark33: Good catch! Thanks :)

spark33: Do you have suggestions for what might be better than `Information on LeafyGreen UI ${componentName} component`? That's not a change I introduced, but I'm open to changing it here now

components/SignIn/DesktopSignIn.tsx
spark33: I think this is a bug with the Menu component itself - I'm using `FocusableMenuItem` as designed I believe

pages/api/auth/api-auth.ts
TheSonOfThomp: 👍 Should we create a ticket (or at least a `// TODO`) to update this once we're on R18?

utils/ContentStack/getContentstackResources.ts
TheSonOfThomp: When it's shown next to a checkbox, `private` makes sense. 
My concern is just that it's ambiguous from looking at this code whether `private` is a boolean, or an object with some private data.
We can leave it, as long as it's documented

spark33: We could change this to an enum so there's Contentstack-side labels and code-side labels as separate strings maybe. I'll do that now

layouts/ComponentLayout.tsx
```diff
@@ -87,6 +87,11 @@ const tabStyles = css`
   }
 `;
 
+const GH_ORGS = {
```
TheSonOfThomp, line 90: nit: abbreviation could be a bit unclear. `GITHUB_ORGS` isn't that much longer

layouts/ComponentLayout.tsx
```diff
@@ -102,7 +107,7 @@ const ComponentLinks = ({
       rel="noopener noreferrer"
       style={{ marginRight: '8px' }}
       href={`https://github.com/${
-        component.private ? '10gen' : 'mongodb'
+        component.private ? GH_ORGS.private : GH_ORGS.public
```
TheSonOfThomp, line 110: Can we break this out into a utility? `getGitHubLink(component)`

pages/api/auth/api-auth.ts
spark33: Good idea! https://jira.mongodb.org/browse/LG-3702

utils/ContentStack/getContentstackResources.ts
spark33: Actually, maybe we should be maintaining parity with Contentstack in terms of naming so it doesn't get confusing? At the very least, are we okay with sitting with this for now if we agree `private` makes more sense on Contentstack? If we know future changes will be purely dev-side, we don't have to do much.

layouts/ComponentLayout.tsx
TheSonOfThomp: Just played around with VoiceOver. This should be fine

TheSonOfThomp (approved):
LGTM 👍

components/SignIn/DesktopSignIn.tsx
shaneeza: might be worth taking a look at what [MongoNav](https://github.com/10gen/leafygreen-ui-private/blob/main/packages/mongo-nav/src/org-nav/UserMenu/UserMenu.tsx#L406) does since it also uses `FocusableMenuItem`

PR #307, opened 2023-10-16
LG-3691: Update faker version to fix table live example

spark33:


yarn.lock
```diff
@@ -1902,10 +1902,10 @@
   resolved "https://registry.yarnpkg.com/@eslint/js/-/js-8.50.0.tgz#9e93b850f0f3fa35f5fa59adfd03adae8488e484"
   integrity sha512-NCC3zz2+nvYd+Ckfh87rA47zfu2QsQpvc6k1yzTk+b9KzRj0wkGa8LSoGOXN6Zv4lRf/EIoZ80biDh9HOI+RNQ==
 
-"@faker-js/faker@^7.6.0":
-  version "7.6.0"
-  resolved "https://registry.yarnpkg.com/@faker-js/faker/-/faker-7.6.0.tgz#9ea331766084288634a9247fcd8b84f16ff4ba07"
-  integrity sha512-XK6BTq1NDMo9Xqw/YkYyGjSsg44fbNwYRx7QK2CuoQgyy+f1rrTDHoExVM5PsyXCtfl2vs2vVJ0MN0yN6LppRw==
+"@faker-js/faker@^8.0.0":
+  version "8.2.0"
```
TheSonOfThomp, line 1906: Should we make sure the resolved version is the same?

TheSonOfThomp (approved):

yarn.lock
spark33: I feel like it should be alright since the biggest thing here is the API change? I think it would be more important if all of our LGUI components used the same version of faker, but modal is 8.0.2 and table is on 8.0.0 and it feels like patch/minor versions can be disregarded

spark33: If this becomes an issue, we can lock it down

PR #333, opened 2023-12-05
LG-3814: MongoNav LiveExample + CodeDocs

spark33:


.npmrc
```diff
@@ -1,2 +1,4 @@
 @lg-private:registry=https://artifactory.corp.mongodb.com/artifactory/api/npm/leafygreen-ui/
-//artifactory.corp.mongodb.com/artifactory/api/npm/leafygreen-ui/
\ No newline at end of file
+//artifactory.corp.mongodb.com/artifactory/api/npm/leafygreen-ui/:username=design-systems-team
```
bruugey, line 2: this shouldn't need to be committed. move to your local  .npmrc file instead at ~/

components/pages/example/LiveExample.styles.ts
```diff
@@ -7,7 +7,8 @@ export const liveExampleWrapperStyle = css`
   position: relative;
   display: flex;
   margin-bottom: initial;
-  overflow: hidden;
+  // 'overflow: hidden' will stop MongoNav's Menu's from being hidden
```
bruugey, line 10: rm?

spark33: Since this is a change that will be applied to all stories, this is just a comment to ensure we don't fix some other issue by adding `overflow: hidden`. I did my best to check all the stories, but just in case there's something I missed.

package.json
```diff
@@ -88,6 +87,7 @@
     "@leafygreen-ui/tokens": "^2.2.0",
     "@leafygreen-ui/tooltip": "^11.0.0",
     "@leafygreen-ui/typography": "^18.0.1",
+    "@lg-private/mongo-nav": "^13.1.2",
```
TheSonOfThomp, line 90: bump

components/pages/example/KnobsTable/Knob/Knob.tsx
```diff
@@ -92,6 +92,7 @@ export const Knob = ({
             value={value}
             onChange={onChange}
             className={cx(inputStyle, rest.className)}
+            popoverZIndex={999}
```
TheSonOfThomp, line 95: 🚩Do we need this?

utils/_getComponentResources.tsx
```diff
@@ -46,10 +48,16 @@ export async function getChangelog(
   componentName: string,
 ): Promise<string | null> {
   try {
+    let namespace = '@leafygreen-ui';
+
+    if (PRIVATE_PACKAGES.includes(componentName)) {
+      namespace = '@lg-private';
+    }
```
TheSonOfThomp, line 55: `const namespace = PRIVATE_PACKAGES.includes(componentName) ? '@lg-private' : '@leafygreen-ui'`

utils/_getComponentResources.tsx
```diff
@@ -62,10 +70,16 @@ export async function getChangelog(
 
 export async function getReadme(componentName: string): Promise<string | null> {
   try {
+    let namespace = '@leafygreen-ui';
+
+    if (PRIVATE_PACKAGES.includes(componentName)) {
+      namespace = '@lg-private';
+    }
```
TheSonOfThomp, line 77: same ternary here

utils/_getComponentResources.tsx
```diff
@@ -81,14 +95,19 @@ export async function getReadme(componentName: string): Promise<string | null> {
 export async function getTSDoc(
   componentName: string,
 ): Promise<Array<CustomComponentDoc> | null> {
+  let namespace = '@leafygreen-ui';
   if (typeof componentName !== 'string') return null;
 
+  if (PRIVATE_PACKAGES.includes(componentName)) {
+    namespace = '@lg-private';
+  }
+
```
TheSonOfThomp, line 104: ternary here as well

utils/getComponentStories.ts
```diff
@@ -12,8 +14,14 @@ export type ModuleType = {
 export async function getComponentStories(
   kebabName: string,
 ): Promise<ModuleType | undefined> {
+  let namespace = '@leafygreen-ui';
+
+  if (PRIVATE_PACKAGES.includes(kebabName)) {
+    namespace = '@lg-private';
+  }
+
```
TheSonOfThomp, line 22: ternary here too

TheSonOfThomp: (could also create a `getNamespaceFromPackageName` function for future extensibility)

TheSonOfThomp (commented):
Left some comments

utils/_getComponentResources.tsx
spark33: I wrote this this way since if we were to include `lg-chat` in the future, we'd want `if / else if`

utils/getComponentStories.ts
spark33: ^I like that

spark33:
Seeing a 1yr+ old PR of mine merged wasn't how I expected to start my morning - feels like I never left! 🤣🚀

PR #334, opened 2023-12-15
Adds Codesandbox links to components!

bruugey:
- Updates existing logo icons
- Fixes annoying border


layouts/ComponentLayout.tsx
```diff
@@ -78,10 +77,16 @@ const desktopLinksContainer = css`
 const tabStyles = css`
   width: 100%;
   max-width: 100%;
+
```
shaneeza, line 80: I think sean fixed the uneven border in the tabs component, are we using the latest?

shaneeza: https://github.com/mongodb/leafygreen-ui/pull/2040

bruugey: is staging outof sync from main?

shaneeza: `Tabs` is at `v11.1.13` which is correct, maybe the fix didn't work?

utils/ContentStack/types.ts
```diff
@@ -31,6 +31,7 @@ export interface ComponentPageMeta {
   url: string;
   description: string;
   figmaurl?: string;
+  codesandbox_url?: { title?: string; href?: string };
```
shaneeza, line 34: whats the `title` for?

bruugey: it's just how the data is stored in contentstack

TheSonOfThomp: Can we change that on the CS side to match the figmaurl schema?

bruugey: Figma url was aded as a text field, this is technically more correct - but I can update for consistency if you all prefer.

layouts/ComponentLayout.tsx
```diff
@@ -108,6 +123,7 @@ const ComponentLinks = ({
     </IconButton>
     {component.figmaurl && (
       <IconButton
+        key="figma"
```
TheSonOfThomp, line 126: do we need a `key` in an un-mapped element?

bruugey: was requesting one

TheSonOfThomp (approved):

PR #337, opened 2023-12-19
Updates icons to tooltips

bruugey:
Also fixes component display names

### DIFF
--- a/layouts/ComponentLayout.tsx
+++ b/layouts/ComponentLayout.tsx
@@ -23,6 +23,7 @@ import LeafyGreenProvider from '@leafygreen-ui/leafygreen-provider';
 import { palette } from '@leafygreen-ui/palette';
 import { Tab, Tabs } from '@leafygreen-ui/tabs';
 import { breakpoints, spacing } from '@leafygreen-ui/tokens';
+import Tooltip from '@leafygreen-ui/tooltip';
 import { H2 } from '@leafygreen-ui/typography';
 
 const layout = css`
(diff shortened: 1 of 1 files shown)


layouts/ComponentLayout.tsx
```diff
@@ -89,6 +90,74 @@ const tabStyles = css`
...
+  if (!href) {
+    return null;
+  }
+
+  return (
+    <Tooltip
+      trigger={
+        <IconButton
+          key={type}
+          aria-label="View in CodeSandbox"
```
shaneeza, line 145: Needs to be updated

layouts/ComponentLayout.tsx
```diff
@@ -89,6 +90,74 @@ const tabStyles = css`
...
+          as="a"
+          target="_blank"
+          rel="noopener noreferrer"
+          style={{ marginRight: '8px' }}
```
shaneeza, line 149: Do we want margin-right on all of these?

bruugey: that's what we had before

layouts/ComponentLayout.tsx
bruugey: fixed

layouts/ComponentLayout.tsx
spark33: It's a nit, but it would be better to use `gap: ${spacing[1]}px` now that there are 3 elements

spark33 (approved):

PR #339, opened 2024-01-01
fix: extract default critical styles to prevent FOUC

liao-frank:
The only critical styles currently being extracted during SSR are the ones used for the leafygreen component library, causing FOUC:

https://github.com/mongodb/design/blob/b10912da367d3d456f4a8cbc97bd700711caf633/utils/renderer.ts#L1
https://github.com/mongodb/design/blob/b10912da367d3d456f4a8cbc97bd700711caf633/utils/renderer.ts#L12-L14

localhost_3000_ (1)

As long as the `css` function from `@emotion/css` is being used for the docs site, then also need to extract the default critical styles to prevent FOUC. Unhydrated content when extracting all critical styles:

localhost_3000_

Alternatively, could migrate to using the `css` function from `@leafygreen-ui/emotion`, or could migrate to using the `css` function from `@emotion/react` and then use `CacheProvider` to provide the custom leafygreen cache to the tree. Both alternatives would mean "coupling" the doc site styling to the leafygreen library, which might be undesirable.

Whatever the right solution is, it should probably also be documented in the README for [@leafygreen-ui/emotion](https://www.npmjs.com/package/@leafygreen-ui/emotion?activeTab=readme). Some users of leafygreen might want to keep multiple pieces of extraction functionality since using multiple caches has legitimate purposes. 

Quote from current README:

> NOTE: If you are already server-side rendering an application using Emotion, you will use the methods exposed in @leafygreenui/emotion instead of, NOT in addition to the methods exposed by emotion-server.


pages/_document.tsx
```diff
@@ -14,17 +14,21 @@ import { renderStatic } from 'utils/renderer';
...
+    const { styles } = renderStatic(() => page.html);
     const initialProps = await Document.getInitialProps(ctx);
+
     return {
       ...initialProps,
       styles: (
         <>
           {initialProps.styles}
-          <style
-            data-emotion={`css ${ids.join(' ')}`}
```
liao-frank, line 25: Note: The `css` key here is wrong regardless of this PR.

spark33:
Thanks for the PR @liao-frank! This looks awesome.

For internal users - this fixes [LG-3141](https://jira.mongodb.org/browse/LG-3141).

PR #361, opened 2024-05-20
Fix focus visible and links on component cards

bruugey:


src/components/global/SideNavigation.tsx
```diff
@@ -187,8 +190,11 @@ export function SideNavigation() {
           href={'/'}
         >
           <MongoDBLogo
+            key="logo"
             height={24}
-            color={darkMode ? SupportedColors.White : SupportedColors.Black}
+            color={
+              darkMode ? SupportedColors.White : SupportedColors.GreenDark2
+            }
```
stephl3, line 197: did design request switching back to green? it was switched to black because of a request from rob

bruugey: he did but then he switched his mind back haha

stephl3 (approved):

PR #369, opened 2024-05-29
Refactor session logic into its own hook

bruugey:
- Adds `useSession` hook
- Creates index file for all hooks
- Changes hooks to named exports from default, and imports them from hooks root
- Moves `700px` max width to the right place (currently not taking effect)

### DIFF
--- a/src/components/global/UserMenu.tsx
+++ b/src/components/global/UserMenu.tsx
@@ -1,6 +1,5 @@
 'use client';
 
-import { useEffect, useState } from 'react';
 import { css } from '@emotion/css';
 import Button from '@leafygreen-ui/button';
 // @ts-expect-error
--- a/src/components/global/UserMenu.tsx
+++ b/src/components/global/UserMenu.tsx
@@ -1,27 +1,20 @@
-"use client";
+'use client';
 
-import { useEffect, useState } from "react";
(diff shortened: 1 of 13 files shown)


src/components/global/UserMenu.tsx
```diff
@@ -1,27 +1,20 @@
-"use client";
+'use client';
 
-import { useEffect, useState } from "react";
-import { css } from "@emotion/css";
-import Button from "@leafygreen-ui/button";
+import { useEffect, useState } from 'react';
```
stephl3, line 3: think we can drop these imports at this point?

src/hooks/useSession.ts
```diff
@@ -0,0 +1,18 @@
+'use client';
+
+import { useEffect, useState } from 'react';
+import { getSession, Session } from '@/auth';
+
+export const useSession = () => {
```
stephl3, line 6: think we can also add it to this file: https://github.com/mongodb/design/blob/staging/src/app/private/page.tsx#L16-L24

src/components/global/UserMenu.tsx
bruugey: Yep good catch

stephl3 (approved):

PR #370, opened 2024-06-10
Adds Icon page back to .design

bruugey:
- Updates to use latest version of all LG components
- Max width for design docs pages is 700px


stephl3 (commented):
understandable if we opt not to prioritize mobile breakpoints but should we still have responsiveness until the 1024px?
Screenshot 2024-06-10 at 3 19 32 PM

seems like a solution could be to update the `grid-template-columns` value in the `Icon.stories.tsx` file to `repeat(4, 1fr)`. at that point, it could also make sense to add a fixed-height w/ scroll 
Screenshot 2024-06-10 at 3 20 01 PM

stephl3 (approved):

PR #374, opened 2024-07-08
Updates components to latest

bruugey:


src/components/global/Search/Search.tsx
```diff
@@ -54,34 +59,19 @@ export function Search() {
...
       onChange={handleSearchChange}
-      className={css`
-        margin: ${spacing[400]}px ${spacing[600]}px;
-      `}
+      className={searchInputStyle}
     >
       {results.map(item => (
         <SearchResult
           key={item.name}
+          onClick={() => router.push(item.navPath ?? '/')}
```
shaneeza, line 67: I'm curious: Why did you change this?

bruugey: Link broke the styles - which is why that needs to be major version bumped.

shaneeza (approved):

PR #411, opened 2025-02-07
Adds back missing live examples from Typography, Palette 

bruugey:


src/utils/patterns.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const titleCase = (component: Pattern) => {
+  return startCase(toLower(component));
+};
+
+const generateComponentNavPath = (component: Pattern) =>
+  `/component/${component}/live-example`;
+
+const generatePatternPath = (foundation: Pattern) =>
+  `/foundation/${foundation}`;
```
shaneeza, line 19: Is there a utils file/directory you can add these to since you're repeating them in the file above?

shaneeza (commented):
was reverting the nav back to the old layout intentional?

bruugey:
> was reverting the nav back to the old layout intentional?

What do you mean?

bruugey:
> > was reverting the nav back to the old layout intentional?
> 
> What do you mean?

If you're talking about the IA, then yes

PR #417, opened 2025-03-10
LG-4754: End of Life page

shaneeza:
[LG-4754](https://jira.mongodb.org/browse/LG-4754)

Screenshot 2025-03-10 at 12 01 19 PM


src/utils/titleCase.ts
shaneeza, line 1: This is to ensure the correct title appears in the side nav.

PR #418, opened 2025-03-14
LG-4966: adds Drawer page and updates radio knob

stephl3:
Ticket: [LG-4966](https://jira.mongodb.org/browse/LG-4966)

- updates LG deps
- adds `Drawer` to side nav
- updates `Knob` with `knobType="radio"` so that deselect is not allowed to avoid errors
  - see selecting/deselecting `variant` in `Banner` live example page
  - see deselecting `language` in `Code` live example page

### DIFF
--- a/src/app/component/[component]/live-example/page.tsx
+++ b/src/app/component/[component]/live-example/page.tsx
@@ -88,21 +88,17 @@ export default function Page({ params }: { params: { component: string } }) {
           border-radius: ${borderRadius[600]}px ${borderRadius[600]}px 0 0;
           background-color: ${color[componentProps?.darkMode ? 'dark' : 'light']
             .background.primary.default};
+          padding: ${spacing[600]}px ${spacing[600]}px 0;
         `}
       >
-        <div
-          className={css`
-            padding-left: ${spacing[600]}px;
-            padding-right: ${spacing[600]}px;
-          `}
-        >
+        <div>
           {/* @ts-expect-error */}
           <LiveExample {...componentProps} />
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


src/components/live-example/Knob/Knob.tsx
```diff
@@ -69,22 +69,44 @@ export const Knob = ({
...
-    case "radio": {
+    case "select": {
       if (knobOptions && knobOptions.length) {
         return (
-          (<Select
-            value={value}
-            onChange={onChange}
+          <Select
+            allowDeselect={true}
+            aria-label={propName}
```
shaneeza, line 77: omg thank you for adding this

shaneeza (commented):
One tiny thing I noticed is that the bottom of the drawer does not align with the bottom of the div

Screenshot 2025-03-17 at 12 46 42 PM

shaneeza:
> > One tiny thing I noticed is that the bottom of the drawer does not align with the bottom of the div
> > Screenshot 2025-03-17 at 12 46 42 PM
> 
> I think it's actually the `border-top` from the `KnobRow` that is overlapping with the story itself. It looks like this is our tallest story currently. I updated the padding of the `/live-example` page so they don't overlap
> 
> Screenshot 2025-03-17 at 10 23 15 AM Screenshot 2025-03-17 at 10 23 06 AM Screenshot 2025-03-17 at 10 23 10 AM

Is it possible to adding the padding to the top div instead? There are some instances, like `Code` where it could use some top and bottom padding.

shaneeza (approved):

